PR #1, opened 2019-08-19
WIP: Add ADR

svtkachenko:


adr/0002-resolve-schema-references-lazily.md
```diff
@@ -0,0 +1,19 @@
...
+
+We need
+
+## Decision
+
+We will
+
+## Consequences
+
+What becomes easier or more difficult to do and any risks introduced by the change that will need to be mitigated.
```
Stranger6667: Test collection phase becomes faster, also it allows to avoid unnecessary schema resolving if certain tests are de-selected.

adr/0002-resolve-schema-references-lazily.md
```diff
@@ -0,0 +1,19 @@
...
+
+Date: 2019-08-08
+
+## Status
+
+Accepted
+
+## Context
+
+We need
```
Stranger6667: We need to have minimal performance overhead on a test collection and avoid resolving that involves network requests

adr/0003-delay-schema-loading.md
```diff
@@ -0,0 +1,19 @@
...
+
+The issue motivating this decision, and any context that influences or constrains the decision.
+
+## Decision
+
+The change that we're proposing or have agreed to implement.
+
+## Consequences
+
+What becomes easier or more difficult to do and any risks introduced by the change that will need to be mitigated.
```
Stranger6667: Basically the same as the AD above - schema could be loaded over the wire, we need to avoid it

adr/0005-use-poetry.md
```diff
@@ -0,0 +1,19 @@
...
+
+The issue motivating this decision, and any context that influences or constrains the decision.
+
+## Decision
+
+The change that we're proposing or have agreed to implement.
+
+## Consequences
+
+What becomes easier or more difficult to do and any risks introduced by the change that will need to be mitigated.
```
Stranger6667: Using PEP 518 simplifies distribution - we have only `pyproject.toml`

PR #4, opened 2019-08-21
Dereferencing draft

svtkachenko:


src/schemathesis/generator.py
```diff
@@ -46,5 +46,6 @@ def get_parameters_strategy(parameters: ParametersList) -> st.SearchStrategy:
 def get_strategy(item: Dict[str, Any]) -> st.SearchStrategy:
     if "schema" in item:
         item = item["schema"]
-    item.pop("required", None)
+    if not isinstance(item.get("required"), list):
```
Stranger6667: Please, add a test for this block

test/schemas/test_parametrization.py
```diff
@@ -133,3 +133,44 @@ def test_(request, case):
     result = testdir.runpytest("-v", "-s")
     result.assert_outcomes(passed=1)
     result.stdout.re_match_lines([r"Hypothesis calls: 1"])
+
+
+def test_recursive_dereference(testdir):
+    # When a given parameter contains a JSON reference
```
Stranger6667: "..., that reference an object with another reference"

test/schemas/test_parametrization.py
```diff
@@ -133,3 +133,44 @@ def test_(request, case):
...
+        },
+    )
+    # Then it should be correctly resolved and used in the generated case
+    result = testdir.runpytest("-v", "-s")
+    result.assert_outcomes(passed=1)
+    result.stdout.re_match_lines([r"Hypothesis calls: 1"])
+
+
+def test_custom_func(testdir):
+    # When a given parameter contains a JSON reference
```
Stranger6667: "When the given string parameter has a custom `format` value"

test/schemas/test_parametrization.py
```diff
@@ -133,3 +133,44 @@ def test_(request, case):
...
+        """
+@schema.parametrize(max_examples=1)
+def test_(request, case):
+    request.config.HYPOTHESIS_CASES += 1
+    assert case.path == "/v1/users"
+    assert case.method == "GET"
+""",
+        **as_param({"parameter": {"type": "string", "format": "custom_format", "in": "query"}}),
+    )
+    # Then it should be correctly resolved and used in the generated case
```
Stranger6667: There could be two comments for different assertions
"Then the relevant test case should be skipped"

"And a proper message is written to the output"

src/schemathesis/schemas.py
```diff
@@ -63,24 +65,44 @@ def get_full_path(self, path: str) -> str:
...
         self, filter_method: Optional[Filter] = None, filter_endpoint: Optional[Filter] = None
     ) -> Generator[Endpoint, None, None]:
-        paths = self.raw_schema["paths"]  # pylint: disable=unsubscriptable-object
+        schema = deepcopy(self.raw_schema)  # modifications are going to happen
+        paths = schema["paths"]  # pylint: disable=unsubscriptable-object
         for path, methods in paths.items():
             full_path = self.get_full_path(path)
             if should_skip_endpoint(full_path, filter_endpoint):
                 continue
+            common_parameters = methods.pop("parameters", [])
```
Stranger6667: Pls, add a test for this case

src/schemathesis/schemas.py
```diff
@@ -63,24 +65,44 @@ def get_full_path(self, path: str) -> str:
...
+                # references should be resolved here to know where to put the parameter - body / query / etc
                 parameters = [self.prepare_item(item) for item in parameters]
                 yield Endpoint(path=full_path, method=method.upper(), parameters=parameters)
 
     def prepare_item(self, item: Dict[str, Any]) -> Dict[str, Any]:
         if is_reference(item):
-            item = self.dereference(item["$ref"])
+            item = self.resolve_reference(item["$ref"])
+        elif item["in"] == "body" and is_reference(item["schema"]):
+            # TODO. is it ok to modify it? or better to copy it?
```
Stranger6667: The comment is not relevant anymore

src/schemathesis/schemas.py
```diff
@@ -63,24 +65,44 @@ def get_full_path(self, path: str) -> str:
...
+            item.update(self.resolve_reference(item["schema"]["$ref"]))
+            del item["schema"]
```
Stranger6667: Hm, it seems that I added a bug here.

If this one will be deleted, then if another test under the same `parametrize` call will need this parameter somehow (not sure if it is the case, tho), then this method will fail on the `elif` line above, because `schema` is absent in the dictionary. Maybe it is the case for common parameters, that are reused among multiple methods

src/schemathesis/schemas.py
```diff
@@ -122,3 +144,13 @@ def is_reference(item: Dict[str, Any]) -> bool:
 
 def filter_parameters(parameters: ParametersList, place: str) -> ParametersList:
     return [parameter for parameter in parameters if parameter["in"] == place]
+
+
+def traverse_schema(schema):
```
Stranger6667: Could you, pls, add tests for this function?

And also types

test/schemas/test_petstore.py
```diff
@@ -0,0 +1,8 @@
+# from .utils import get_schema
```
Stranger6667: This one is better to rewrite with `pytester` instead

Stranger6667:
Please, add types for the new code

test/schemas/test_parametrization.py
```diff
@@ -133,3 +133,51 @@ def test_(request, case):
...
+
+
+def test_custom_format(testdir):
+    # When the given string parameter has a custom format value
+    testdir.make_test(
+        """
+@schema.parametrize(max_examples=1)
+def test_(request, case):
+    request.config.HYPOTHESIS_CASES += 1
+    assert case.path == "/v1/users"
```
Stranger6667: These two lines with `assert`s could be removed

test/schemas/test_parametrization.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 
-from .utils import as_param, integer
+from .utils import array, as_param, integer
```
Stranger6667: `array` is not used and could be removed

test/schemas/test_parametrization.py
```diff
@@ -133,3 +133,66 @@ def test_(request, case):
...
+def test_(request, case):
+    request.config.HYPOTHESIS_CASES += 1
+""",
+        **as_param({"name": "parameter", "type": "string", "format": "custom_format", "in": "query"}),
+    )
+    result = testdir.runpytest("-v", "-rs")
+    # Then the relevant test case should be skipped
+    result.assert_outcomes(skipped=1)
+    # And a proper message is written to the output
+    result.stdout.re_match_lines([".* Unsupported string format=custom_format"])
```
Stranger6667: The full message includes some part of the traceback, right? could we remove the traceback and leave only the error message?

Stranger6667: Not relevant

PR #13, opened 2019-08-26
Convert parameters list to schema

svtkachenko:


src/schemathesis/schemas.py
```diff
@@ -159,8 +159,18 @@ def is_reference(item: Dict[str, Any]) -> bool:
...
+def convert_parameters(parameters: List[Dict[str, Any]], place: str) -> Dict[str, Any]:
+    new: Dict[str, Any] = {"properties": {}, "additionalProperties": False, "type": "object"}
+    for parameter in parameters:
+        if parameter["in"] == place:
+            if parameter.get("required", False):
+                new.setdefault("required", [])
+                new["required"].append(parameter["name"])
+            new["properties"][parameter["name"]] = {}
+            for key, value in parameter.items():
+                if key not in ("required", "name", "in"):
```
Stranger6667: Could you, pls, add a comment that these keys are not supported by JSON schema?

src/schemathesis/schemas.py
```diff
@@ -159,8 +159,18 @@ def is_reference(item: Dict[str, Any]) -> bool:
...
+                    new["properties"][parameter["name"]][key] = value
```
Stranger6667: The left part is a bit long and has some nesting that we can avoid:

```
def convert_property(data):
    new_property = {}
    for key, value in parameter.items():
        if key not in ("required", "name", "in"):
            new_property[key] = value
    return new_property
```

then:

```
new["properties"][parameter["name"]] = convert_property(parameter)
```

test/schemas/types/test_complex.py
```diff
@@ -23,9 +35,9 @@ def test_(case):
         """.format(
             name=parameter["name"]
         ),
-        **as_array(items=parameter),
+        **as_array(items=parameter, required=True),
     )
-    testdir.run_and_assert(passed=1)
```
Stranger6667: This line might be not needed

src/schemathesis/schemas.py
```diff
@@ -159,8 +159,18 @@ def is_reference(item: Dict[str, Any]) -> bool:
     return "$ref" in item
 
 
-def filter_parameters(parameters: ParametersList, place: str) -> ParametersList:
-    return [parameter for parameter in parameters if parameter["in"] == place]
+def convert_parameters(parameters: List[Dict[str, Any]], place: str) -> Dict[str, Any]:
```
Stranger6667: Add a docstring pls do describe why is it needed

src/schemathesis/schemas.py
```diff
@@ -27,16 +27,16 @@ class Endpoint:
...
     @property
-    def query(self) -> ParametersList:
-        return filter_parameters(self.parameters, "query")
+    def query(self) -> Dict:
+        return convert_parameters(self.parameters, "query")
 
     @property
-    def body(self) -> ParametersList:
-        return filter_parameters(self.parameters, "body")
+    def body(self) -> Dict:
```
Stranger6667: `Dict[str, Any]`

PR #14, opened 2019-08-29
Update README and docs

svtkachenko:


README.md
```diff
@@ -15,16 +17,23 @@ To generate test cases for your schema you need:
...
 
 schema = Parametrizer.from_path("path/to/schema.yaml")
 
+@pytest.fixture(scope="session")
+def client():
+    return requests.Session()
 
 @schema.parametrize()
 def test_users_endpoint(client, case):
+    url = "http://0.0.0.0:8080" + case.formatted_path
```
Stranger6667: this `url` should be used below

README.md
```diff
@@ -15,26 +17,34 @@ To generate test cases for your schema you need:
...
+import pytest
+import requests
 from schemathesis import Parametrizer
 
 
 schema = Parametrizer.from_path("path/to/schema.yaml")
 
+@pytest.fixture(scope="session")
+def client():
+    return requests.Session()
```
Stranger6667: it would be nice to have it like this:

```
with requests.Session() as session:
    yield session
```

so the session is closed in the end

Issue #26, opened 2019-09-09
Generate test cases from the given parameters


Stranger6667:
We need to create N instances of `Case` from the given parameters, it could be done by creating a function wrapped into `given` call that will add all cases into a list from the higher scope. Code snippet to illustrate the idea:

 
```
def generate_cases(*args, **kwargs): 
    data = [] 
    @given(i=st.integers()) 
    def generator(i): 
        data.append(i) 
    generator() 
    return data
```

Or it could be a method on `Parametrizer` for example:

 
```
class Parametrizer:
    ...
    def generate_cases(self, **kwargs):
        ...
```
Pytest plugin generates different tests for endpoints / methods combos and then wrap them into given - maybe this logic could be factored out from the plugin itself, not sure, but we need to gather cases from all these combos

Stranger6667:
Done via runner #28

PR #36, opened 2019-09-11
Catch exceptions in hasattr checks

orsinium:
When using the project with Flask, it raises `RuntimeError` on `hasattr` check. I think, `is_schemathesis_test` check should be stable against all exceptions in `__getattr__` of the test function because it means that `_schema_parametrizer` wasn't explicitly specified by Schemathesis. That's not the last problem with it but step towards to integration of the project with uninitialized Flask app.

### DIFF
--- a/src/schemathesis/parametrizer.py
+++ b/src/schemathesis/parametrizer.py
@@ -92,4 +92,7 @@ class SchemaWrapper:
 
 def is_schemathesis_test(func: Callable) -> bool:
     """Check whether test is parametrized with schemathesis."""
-    return hasattr(func, "_schema_parametrizer")
+    try:
+        return hasattr(func, "_schema_parametrizer")
+    except Exception:
+        return False


Stranger6667:
Hej!
Thank you for submitting a patch :) What would be the minimal failing test case here? So we can add a regression test

Stranger6667:
Btw, I'd be happy to help with fixing the rest on problems :) Let me know if I can do something

Stranger6667:
Will add a failing test case, changelog entry and will make a new release soon

Stranger6667:
`0.4.1` is released

orsinium:
> Btw, I'd be happy to help with fixing the rest on problems :) Let me know if I can do something

I've tried to adapt it for our project, but haven't succeeded. We have no existing file with the specification, only [apispec](https://apispec.readthedocs.io/en/stable/). We can generate docs from it,  but:

1. Schemathesis must generate test cases BEFORE running any tests, on the stage of tests collecting.
2. Flask app initialization happens in a fixture, before running the first test.

I thought about laziness for test cases generation but looks like there is no way to do it. Maybe, only some kind of subtests inside existing test :thinking: 

My decision, for now, is keeping schemathesis tests as a separate script, outside of main tests.

Stranger6667:
Interesting use case :)

Currently, the schema is used to generate parametrized tests for <method>-<endpoint> combinations.

Without this part, it should be possible to defer the schema loading more than it is now with `deferred` strategy from `hypothesis`. However, I'm not sure if it will work in the way that `case` will be dependent on some fixture, need to check it.

Re subtests:

We could generate test cases first, and then run them - we want it for CLI runner (#26) and then it should be possible to generate cases and use them in subtests. But in this case, we will not give feedback to the `hypothesis` machinery and shrinking won't work at all, still thinking about how to do it better.

There is also https://github.com/untitaker/pytest-subtesthack but not sure if it could help (maybe it could give some ideas tho)

Maybe something like this will work:

```
def test_api(flask_app):
    schema = ...  # get the schema from the app

    parametrizer = Parametrizer(schema)

    strategies = parametrizer.to_strategies()  # strategies per <method>-<endpoint> combo

    @given(case=strategies[0])
    def test_a(case):
        ...
    # or generate multiple functions dynamically here
    test_a()
```

PR #63, opened 2019-09-23
[WIP] Test executor

Stranger6667:
Initial draft on the logic, that will execute certain checks against some app

TODO:

- [ ] Docs
- [x] Tests should include some locally running server (similar to Django's `live_server`) so we can actually verify the behavior. Probably the simplest option is to use Flask or aiohttp server (aiohttp even has a builtin fixture for it).
- [ ] Collect test execution results
- [x] move `aiohttp` to dev dependencies

In the future, it could be also WSGI apps, so `requests` might be optional. Also, some parts could be done asynchronously (tests itself could be run in parallel if there are no restrictions from the hypothesis side), but I am not sure how to implement it better :(

### DIFF
--- a/src/schemathesis/runner.py
+++ b/src/schemathesis/runner.py
@@ -1,21 +1,35 @@
-from typing import Callable
+from typing import Callable, NoReturn
+from urllib.parse import urlsplit, urlunsplit
 
 import requests
 
+from .loaders import from_uri
 from .models import Case
-from .schemas import BaseSchema
 
 
-def execute(url: str, schema: BaseSchema) -> None:
-    def single_test(case: Case, check: Callable) -> None:
-        response = requests.request(
-            case.method, f"{url}{case.formatted_path}", headers=case.headers, params=case.query, json=case.body
-        )
-        check(response)
-
-    for _, test in schema.get_all_tests(single_test):
-        test(check_1)
+def execute(schema_uri: str) -> NoReturn:
+    with requests.Session() as session:
+        schema = from_uri(schema_uri)
+        base_url = get_base_url(schema_uri)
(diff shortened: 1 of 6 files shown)


src/schemathesis/runner.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def execute(url: str, schema: BaseSchema) -> None:
+    def single_test(case: Case, check: Callable) -> None:
+        response = requests.request(
+            case.method, f"{url}{case.formatted_path}", headers=case.headers, params=case.query, json=case.body
+        )
+        check(response)
+
+    for _, test in schema.get_all_tests(single_test):
```
paveldedik: Maybe this part would be much faster if we used `async`. Instead of using `requests`, we could maybe use `aiohttp.ClientSession`?

Stranger6667: My initial try to implement it with aiohttp failed, unfortunately :( It would be nice, yep.

The problem was, that the test itself is sync (however there are some ways to fix it via hypothesis hooks - https://hypothesis.readthedocs.io/en/latest/details.html?highlight=executor#custom-function-execution)
which means that there are multiple async tasks that execute blocking code which doesn't bring benefit - thread/process pool can fix this, but probably it would be better to go with non-blocking tasks (that executor thing above).

I'd be very happy to go in this direction and appreciate any help :)

paveldedik: I see. I will check it when I have time :-). I don't know async very well unfortunately, but this seems like something that should be possible.

Stranger6667:
Also, it probably will make sense to either remove the `deadline` health check or set it to a higher value.
If we will remove it, then it maybe will help with parallel test execution

PR #79, opened 2019-09-27
Add support for filtering in loaders

### LINKED ISSUE #75: Add support for filtering in loaders
It could be helpful to instantiate a schema like this:

```
# test_users_api.py
schema = schemathesis.from_path("path.yaml", endpoint="/v1/users", method="GET")
...
```
E.g. to separate tests in different modules

svtkachenko:
Closes #75

### DIFF
--- a/src/schemathesis/schemas.py
+++ b/src/schemathesis/schemas.py
@@ -19,6 +19,7 @@ from ._hypothesis import create_test
 from .filters import should_skip_endpoint, should_skip_method
 from .models import Endpoint
 from .types import Filter
+from .utils import NOT_SET
 
 
 @attr.s(slots=True)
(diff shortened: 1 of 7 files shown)


src/schemathesis/schemas.py
```diff
@@ -44,8 +44,12 @@ def get_all_tests(
         for endpoint in self.get_all_endpoints():
             yield endpoint, create_test(endpoint, func, settings)
 
-    def parametrize(self, method: Optional[Filter] = None, endpoint: Optional[Filter] = None) -> Callable:
```
Stranger6667: Please, create a sentinel object, instead of a string and use it as a default everywhere for `method` or `endpoint`:

```
NOT_SET = object()
```

test/conftest.py
```diff
@@ -57,9 +57,9 @@ def maker(content, **kwargs):
         from test.utils import *
         from hypothesis import settings
         raw_schema = {schema}
-        schema = schemathesis.from_dict(raw_schema)
+        schema = schemathesis.from_dict(raw_schema, method={method}, endpoint={endpoint})
         """.format(
-                schema=schema
+                schema=schema, method=repr(method), endpoint=repr(endpoint)
```
Stranger6667: Why is `repr` needed?

svtkachenko: To have `method="GET"` instead of `method=GET` after format.

svtkachenko: And `method=None` instead of `method="None"` for hardcoded quotes.

Stranger6667: I'd suggest using double quotes in the template instead

Stranger6667: Update: in this case `None` will become `"None"`

PR #83, opened 2019-09-30
Update filtration for LazySchema

### LINKED ISSUE #64: Possible conflict in filtration by method / endpoint
for example, there is a lazy schema that contains some filters by method/endpoint (should be created directly, not via a loader) in this case if `method` / `endpoint` is not set in `parametrize` arguments then those filters in the schema will be overridden with `None`, which is not what could be expected. 
Not that relevant at the moment, but could be more important when we'll allow adding filters to loaders

svtkachenko:
Closes #64

### DIFF
--- a/src/schemathesis/lazy.py
+++ b/src/schemathesis/lazy.py
@@ -8,13 +8,14 @@ from pytest_subtests import SubTests
 from .models import Endpoint
 from .schemas import BaseSchema
 from .types import Filter
+from .utils import NOT_SET
 
 
 @attr.s(slots=True)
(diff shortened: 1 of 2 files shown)


src/schemathesis/lazy.py
```diff
@@ -5,6 +5,8 @@
 from _pytest.fixtures import FixtureRequest
 from pytest_subtests import SubTests
 
+from schemathesis.utils import NOT_SET
```
Stranger6667: You could use a relative import here

Issue #30, opened 2019-09-09
CLI


Stranger6667:
A CLI tool that will:

- generate test cases from the given schema (probably with an option to show them in stdout);
- execute these use cases against certain URL (if not specified in the schema);
- Pass certain config options to the relevant flow part (test generation / execution);

Possible examples:

Generate cases: 

```
schemathesis -s https://example.com/api/swagger.json --format=json
```

Execute: 
```
schemathesis run https://example.com/api/swagger.json
```

Stranger6667:
The implementation could be based on #63

paveldedik:
@Stranger6667 I don't understand this - what exactly do we want to generate here? I'm talking about this command:

```
schemathesis -s https://example.com/api/swagger.json --format=json
```

What should be the output? And how will the output be used? Should it be just a JSON file containing generated some data representing different cases (inputs for the test function)?

I want to understand why it would be useful, since the second command would do the same as the first one, only without outputting the cases. It just generates the cases and runs the test function, the only argument needed is the swagger.json schema file.

The second command would be implemented easily using the `runner.execute` function. The first command would be about serializing/deserializing the generated cases, easy - but I want to make sure I not missing something here as I'm not sure about use cases.

Stranger6667:
The initial idea was to be able to run each phase separately, but probably it should be decided later, since now it is not clear how useful it will be, indeed. I think we can start with `run` command and then extend the CLI if needed

paveldedik:
Yep, I agree with you, if there isn't any specific case where it would be useful to have the first command, I would skip it for now.

I will implement the `run` command.

PR #86, opened 2019-09-30
Add CLI tool for running schemathesis test cases

paveldedik:
- Fix: https://github.com/kiwicom/schemathesis/issues/30

### DIFF
--- a/src/schemathesis/commands.py
+++ b/src/schemathesis/commands.py
@@ -1,0 +1,43 @@
+from typing import Iterable
+from urllib.parse import urlparse
+
+import click
+
+from . import runner
+
+CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
+
+DEFAULT_CHECKS_NAMES = tuple(check.__name__ for check in runner.DEFAULT_CHECKS)
+
+
+@click.group(context_settings=CONTEXT_SETTINGS)
+@click.version_option()
+def main() -> None:
+    """Command line tool for testing your web application built with Open API / Swagger specifications."""
+
+
+@main.command(short_help="Perform schemathesis test.")
+@click.argument("schema", type=str)
+@click.option(
+    "--checks",
+    "-c",
+    multiple=True,
+    help="List of checks to run.",
+    type=click.Choice(DEFAULT_CHECKS_NAMES),
+    default=DEFAULT_CHECKS_NAMES,
+)
+def run(schema: str, checks: Iterable[str] = DEFAULT_CHECKS_NAMES) -> None:
+    """Perform schemathesis test against an API specified by SCHEMA.
+
+    SCHEMA must be a valid URL pointing to an Open API / Swagger specification.
+    """
+    if not urlparse(schema).netloc:
+        raise click.UsageError("Invalid SCHEMA, must be a valid URL.")
+
+    selected_checks = tuple(check for check in runner.DEFAULT_CHECKS if check.__name__ in checks)
+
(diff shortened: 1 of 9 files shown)


Stranger6667:
Thanks! We are going to fix the tests in master soon (caused by #83 )

Stranger6667:
The current master should be fixed now (GH actions are stuck for some reason)

src/schemathesis/commands.py
```diff
@@ -0,0 +1,56 @@
...
+    help="List of checks to run.",
+    type=click.Choice(default_checks),
+    default=default_checks,
+)
+def run(schema: str, url: str = "", checks: Iterable[str] = default_checks) -> None:
+    """Perform schemathesis test against an API specified by SCHEMA.
+
+    SCHEMA must be a valid URL or a file path.
+    """
+    is_file_path = re.search(r"^https?://", schema) is None
```
Stranger6667: I am not currently sure about a file-based schema in CLI and would rather postpone it a bit (in a separate PR probably) to start with a minimal CLI. What do you think?

src/schemathesis/models.py
```diff
@@ -30,7 +30,7 @@ def empty_object() -> Dict[str, Any]:
 
 
 @attr.s(slots=True)
-class Endpoint:
+class Endpoint:  # pylint: disable=too-many-instance-attributes
```
Stranger6667: There is a pylint comment at the beginning of the file. Probably we need to remove it now

src/schemathesis/runner.py
```diff
@@ -3,24 +3,38 @@
...
 
 
 def not_a_server_error(response: requests.Response) -> None:
     """A check to verify that the response is not a server-side error."""
     assert response.status_code < 500
 
 
-def execute(schema_uri: str, checks: Iterable[Callable] = (not_a_server_error,)) -> None:
-    """Generate and run test cases against the given API definition."""
+default_checks = (not_a_server_error,)
```
Stranger6667: Should we name constants in upper case?

Stranger6667:
In general LGTM! :) Please, check the comments above. However, it will be nice to know the coverage once the build is fixed

src/schemathesis/commands.py
paveldedik: I implemented it because I wanted to test it locally in another app which didn't have the `swagger.json` URL anywhere. It seemed very useful for testing, I can imagine more use-cases (like testing new schema against production or something)...

But I can remove it from this PR definitely, but since it's implemented, should I? Or do you see a big issue with this?

Stranger6667: Local schemas could be tested with `file://` schema, so there will be two options to do the same thing. I need to think more about it and to avoid waiting I wanted to start with the simpler implementation to unblock implementing custom options, etc. I absolutely agree that this feature is useful, I only not sure what would be the best interface for it yet

paveldedik: OK, makes sense. I will remove the option for now (but keep it in another branch if we get back to it).

That said, I personally don't like having `file://`` in front, as the autocomplete feature in bash/zsh doesn't work for me that way. We could make it more explicit by having an explicit option like `-f` or something. But I also like simple APIs which don't require that and are kind of smart (I think the differentiation between URL and file path is straightforward here if we will require `http(s)` in front of the URL).

paveldedik:
BTW the codecov report is outdated

Stranger6667:
yep, sometimes it is not updated :(  I think it is good to merge

PR #102, opened 2019-10-02
[WIP] Add authorization to runner

### LINKED ISSUE #96: Runner. Provide a way to setup authorization
At the moment there is no way to setup auth in `runner._execute_all_tests`. But it should be configurable. 
We can start with these options:
- basic auth (for CLI we can follow cURL convention `--user` option)
- custom header (we can add again, a cURL convention `--header`) so users can add auth header manually

As the next step, it would be nice to have more flexible code in `runner` via e.g. passing arguments to the `requests.Session` or auth callbacks to solve the case when a token expires (and it should be refreshed)

Stranger6667:
Closes #96 

TODO:

- [x] Changelog (won't do, since there is no release yet with CLI)
- [x] Documentation
- [x] Validation & cases when input is in the wrong format
- [x] Mutmut (probably we need to think about periodic runs or runs on a changed subset of the code )

I took `--auth` format from `httpie`, as the next steps we can add digest auth in the way similar to `httpie`
and `--header` format is from `curl`. also it could be similar to approach with prompting for a password but it is not that important now. What do you think?

### DIFF
--- a/src/schemathesis/commands.py
+++ b/src/schemathesis/commands.py
@@ -1,4 +1,5 @@
-from typing import Iterable
+from contextlib import contextmanager
+from typing import Dict, Generator, Iterable, Optional, Tuple
 from urllib.parse import urlparse
 
 import click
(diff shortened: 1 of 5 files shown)


src/schemathesis/commands.py
```diff
@@ -36,8 +49,27 @@ def run(schema: str, checks: Iterable[str] = DEFAULT_CHECKS_NAMES) -> None:
...
+        auth = tuple(raw_auth.split(":"))
+        return cast(Tuple[str, str], auth)
+    return None
+
+
+def prepare_headers(header: Tuple[str, ...]) -> Dict[str, str]:
+    headers = {}
+    for head in header:
+        key, value = head.split(":")
+        headers[key] = value
```
paveldedik: Maybe `value.rstrip()` to remove the space between `Key: Value`?

Stranger6667: Hm, then it should be `lstrip()`, right? I am thinking if space and other not printable chars could be in the header value or not, so we will not remove the part of the value, what do you think?

paveldedik: Yep, `lstrip`. This is what `curl` does. It removes all white-space characters at the beginning.

So doing:

```
curl http://localhost:5000 -H "Authorization:           Bearer 123   "
```

results in header value for key `Authorization`: `Bearer 123   `

Stranger6667: Cool! I didn't know it :) implemented. Thanks for suggesting it :)

PR #103, opened 2019-10-02
[WIP] Speedup tox

Stranger6667:
The current approach is quite slow. Found this comment - https://github.com/tox-dev/tox/issues/715#issuecomment-455192039

Locally I have this for `py37` env:
- master branch - 87 sec.
- this branch - 50 sec.

However, it tests the local code, the package is not installed :(

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,7 +42,7 @@ jobs:
     - uses: actions/setup-python@v1
       with:
         python-version: '3.6'
-    - run: pip install tox coverage
+    - run: pip install tox coverage poetry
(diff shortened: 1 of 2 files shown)


paveldedik:
> However, it tests the local code, the package is not installed :(

What do you mean by this?

Stranger6667:
The desired flow is to run tests against installed version of the package to avoid packaging issues. but this approach runs it against schemathesis in `src` folder, which is not the perfect case, but faster

Stranger6667:
Not relevant due to this option:

```
poetry install
pytest
```

PR #114, opened 2019-10-04
CLI: Fix test generation if no filters are used

paveldedik:
It was broken if no endpoints/methods were used in CLI as the value for endpoints/methods was `()` which than was passed to `should_skip_method` and `should_skip_endpoint`.

### DIFF
--- a/src/schemathesis/utils.py
+++ b/src/schemathesis/utils.py
@@ -1,6 +1,6 @@
 import warnings
 from functools import wraps
-from typing import Any, Callable, List, Set, Tuple, Union
+from typing import Any, Callable, List, Mapping, Set, Tuple, Union
 
 from .types import Filter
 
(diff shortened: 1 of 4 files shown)


Stranger6667:
Good catch! Thank you :)

src/schemathesis/utils.py
```diff
@@ -30,3 +30,8 @@ def force_tuple(item: Filter) -> Union[List, Set, Tuple]:
     if not isinstance(item, (list, set, tuple)):
         return (item,)
     return item
+
+
+def dict_true_values(**kwargs):
+    """Create dict with given kwargs while skipping items where bool(value) evaluates to False."""
+    return {key: kwargs[key] for key in kwargs if bool(kwargs[key])}
```
Stranger6667: We can avoid extra `__getitem__` by using `.items()`:

```
{key: value for key, value in kwargs.items() if bool(value)}
```

What do you think?

paveldedik: Yep, I though we are creating list of tuples by calling `.items()` but now I'm realizing we are not in python 2 anymore :-D Will fix that.

Issue #29, opened 2019-09-09
Process the execution results


Stranger6667:
Given the list of responses / mapping case-response we need to process it and create some summary from it. We can start with a simple report: 

- Number of successful responses
- Number of failed responses

The output could be a simple dictionary, for now, that could be formatted later into stdout text representation or to some JSON.

We need the summary to measure how many failures are there + if we will provide some additional info (e.g. on what input happened certain error), then it will help to fix the issues itself

Stranger6667:
Currently, the execution will stop on the first failing test, but it should execute all of them

PR #120, opened 2019-10-07
cli: Add execution results with colors

paveldedik:
- Fix https://github.com/kiwicom/schemathesis/issues/29

This is what the output looks like:

```
Running schemathesis test cases ...

------------------------------- FALSIFYING EXAMPLES ----------------------------
Falsifying example: single_test(case=Case(path='/ping', method='GET', path_parameters={}, headers={}, cookies={}, query={}, body={}, form_data={}), session=<requests.sessions.Session at 0x7fd21c90f090>, url='http://localhost:8080', checks=(
  <function schemathesis.runner.not_a_server_error>), stats=StatsCollector(data=defaultdict(<class 'collections.Counter'>, {'not_a_server_error': Counter({'total': 100, 'ok': 98, 'error': 2}))})))
--------------------------------------------------------------------------------

----------------------------------- SUMMARY ------------------------------------
not_a_server_error            98 / 100 passed          FAILED 
--------------------------------------------------------------------------------

Tests failed.
```

Maybe a bit overkill, but with hypothesis randomly printing stuff, I wanted to make it a bit more readable.

### DIFF
--- a/src/schemathesis/runner.py
+++ b/src/schemathesis/runner.py
@@ -1,7 +1,9 @@
+from collections import Counter, defaultdict
 from contextlib import suppress
 from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union
 from urllib.parse import urlsplit, urlunsplit
 
+import attr
 import requests
 from requests.auth import AuthBase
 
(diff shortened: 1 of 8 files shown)


src/schemathesis/runner.py
```diff
@@ -59,11 +85,20 @@ def get_base_url(uri: str) -> str:
...
+
     for check in checks:
-        check(response)
+        check_name = check.__name__
+        try:
+            check(response)
+            stats.increment(check_name)
+        except AssertionError as e:
+            stats.increment(check_name, error=e)
+            raise e
```
Stranger6667: Probably we can suppress it until all checks are executed and then reraise some "group" exception, otherwise checks after the first failing one will not be executed. What do you think?

paveldedik: Yep, makes sense.

Stranger6667:
> Maybe a bit overkill, but with hypothesis randomly printing stuff, I wanted to make it a bit more readable.
I like it very much :) Thank you for your effort. Will check in more details soon

PR #131, opened 2019-10-09
Suppress hypothesis exception about using `example` non-interactively

### LINKED ISSUE #92: Suppress hypothesis exception about using `example` non-interactively
There was a recent change in Hypothesis - https://github.com/HypothesisWorks/hypothesis/commit/ac24b2f4e98f6ae0bf2860a36f8b2d2b41bcb7cd

However, in our case, we generate values for `example` decorator when processing the application schema and the warning probably is not that relevant for us.

It will make the test suite output more clear

svtkachenko:
Closes #92


src/schemathesis/_hypothesis.py
```diff
@@ -1,17 +1,22 @@
...
 from hypothesis._strategies import just
+from hypothesis.errors import NonInteractiveExampleWarning
 from hypothesis_jsonschema import from_schema
 
 from .models import Case, Endpoint
 
 PARAMETERS = frozenset(("path_parameters", "headers", "cookies", "query", "body", "form_data"))
 
 
+filterwarnings("ignore", category=NonInteractiveExampleWarning)
```
Stranger6667: I'd rather go for something not global:

```
with warnings.catch_warnings():
    warnings.simplefilter("ignore", NonInteractiveExampleWarning)
    ...
```

src/schemathesis/_hypothesis.py
```diff
@@ -14,14 +16,16 @@
...
     """Create a Hypothesis test."""
-    strategy = endpoint.as_strategy()
-    wrapped_test = hypothesis.given(case=strategy)(test)
-    original_test = get_original_test(test)
-    if asyncio.iscoroutinefunction(original_test):
-        wrapped_test.hypothesis.inner_test = make_async_test(original_test)  # type: ignore
-    if settings is not None:
-        wrapped_test = settings(wrapped_test)
-    return add_examples(wrapped_test, endpoint)
+    with catch_warnings():
```
Stranger6667: It seems like this part is not needed since it is wrapped in `get_examples`

Stranger6667:
Could you, also, please verify that there are no warnings in tests?

src/schemathesis/_hypothesis.py
Stranger6667: Can we move this `catch_warnings` to `get_examples`?

src/schemathesis/_hypothesis.py
```diff
@@ -14,6 +16,7 @@
 
 def create_test(endpoint: Endpoint, test: Callable, settings: Optional[hypothesis.settings] = None) -> Callable:
     """Create a Hypothesis test."""
+    simplefilter("ignore", NonInteractiveExampleWarning)
```
Stranger6667: Is it needed? If not, please remove it

test/test_hypothesis.py
```diff
@@ -1,6 +1,7 @@
 from unittest.mock import ANY
 
 import pytest
+from hypothesis.errors import NonInteractiveExampleWarning
```
Stranger6667: Seems not used

PR #132, opened 2019-10-09
Add a regression test case for runner stopping on the first failure

### LINKED ISSUE #116: Add a regression test case for runner stopping on the first failure
Fixed here - f862d92479048ce5385bec906dd7e64227b07235

I didn't add a regression test case

svtkachenko:
Closes #116

### DIFF
--- a/test/test_runner.py
+++ b/test/test_runner.py
@@ -16,7 +16,7 @@ def app():
     saved_requests = []
 
     async def schema(request):
-        raw = make_schema(paths={"/pets": {"get": {}}})
+        raw = make_schema(paths={"/pets": {"get": {}}, "/zerror": {"get": {}}})
         content = yaml.dump(raw)
         return web.Response(body=content)
 
@@ -31,7 +31,14 @@ def app():
         return web.Response()
 
     app = web.Application()
-    app.add_routes([web.get("/swagger.yaml", schema), web.get("/v1/users", users), web.get("/v1/pets", pets)])
+    app.add_routes(
+        [
+            web.get("/swagger.yaml", schema),
+            web.get("/v1/users", users),
+            web.get("/v1/zerror", users),
+            web.get("/v1/pets", pets),
+        ]
+    )
     app["saved_requests"] = saved_requests
     app["config"] = {"raise_exception": False}
     return app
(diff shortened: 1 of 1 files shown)


test/test_runner.py
```diff
@@ -75,7 +82,7 @@ def assert_not_request(app, method, path):
 def test_execute(server, app):
     headers = {"Authorization": "Bearer 123"}
     execute(f"http://127.0.0.1:{server['port']}/swagger.yaml", api_options=dict(headers=headers))
-    assert len(app["saved_requests"]) == 2
+    assert len(app["saved_requests"]) >= 2
```
Stranger6667: How many exactly?

svtkachenko: 3

Stranger6667: Please, set the exact number

svtkachenko: Next time `app` would be expanded it would affect these numbers in each test in this module :(

Stranger6667: At least it will be explicit

test/test_runner.py
```diff
@@ -88,19 +95,19 @@ def test_execute_base_url(server, app):
...
     app["config"]["raise_exception"] = True
     stats = execute(f"http://127.0.0.1:{server['port']}/swagger.yaml")
     assert "not_a_server_error" in stats.data
-    assert dict(stats.data["not_a_server_error"]) == {"total": 3, "ok": 1, "error": 2}
+    assert dict(stats.data["not_a_server_error"]) == {"total": 5, "ok": 1, "error": 4}
 
 
 def test_auth(server, app):
     execute(f"http://127.0.0.1:{server['port']}/swagger.yaml", api_options=dict(auth=("test", "test")))
-    assert len(app["saved_requests"]) == 2
```
Stranger6667: Could you, please, set the exact number on this line as well?

test/test_runner.py
```diff
@@ -88,19 +95,19 @@ def test_execute_base_url(server, app):
...
+    assert len(app["saved_requests"]) >= 2
```
Stranger6667: Could you, please, set the exact number on this line as well?

PR #133, opened 2019-10-09
Add HTTP Digest Auth support

### LINKED ISSUE #106: Add HTTP Digest Auth support for CLI & runner
Extra auth method could simplify using Schemathesis.
`HTTPDigestAuth` is available in `requests`

svtkachenko:
Closes #106

### DIFF
--- a/src/schemathesis/cli/__init__.py
+++ b/src/schemathesis/cli/__init__.py
@@ -1,6 +1,7 @@
 from typing import Dict, Iterable, Optional, Tuple
 
 import click
+from requests.auth import HTTPDigestAuth
 
 from .. import runner, utils
 from ..types import Filter
@@ -36,6 +37,13 @@ def main() -> None:
     callback=validators.validate_auth,  # type: ignore
 )
 @click.option(  # type: ignore
+    "--auth-type",
+    "-A",
+    type=click.Choice(["basic", "digest"], case_sensitive=False),
+    default="basic",
+    help="The authentication mechanism to be used. Defaults to 'basic'.",
+)
+@click.option(  # type: ignore
     "--header",
     "-H",
     "headers",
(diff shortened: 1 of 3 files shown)


src/schemathesis/cli/__init__.py
```diff
@@ -37,6 +37,14 @@ def main() -> None:
     type=str,
     callback=validators.validate_auth,  # type: ignore
 )
+@click.option(  # type: ignore
```
Stranger6667: I am tending to `httpie` approach with `--auth-type` with two options `basic` (default) and `digest`. In this case, there will be no ambiguity with having two different auth passed as cmd options - there will be only one auth with type switching. What do you think?

src/schemathesis/cli/validators.py
```diff
@@ -13,10 +14,12 @@ def validate_schema(ctx: click.core.Context, param: click.core.Option, raw_value
 
 def validate_auth(
     ctx: click.core.Context, param: click.core.Option, raw_value: Optional[str]
-) -> Optional[Tuple[str, str]]:
+) -> Optional[Union[Tuple[str, str], HTTPDigestAuth]]:
```
Stranger6667: Maybe we can use `Optional[AuthBase]` and always wrap that tuple into `HTTPBasicAuth` ? types will be simpler

src/schemathesis/runner.py
```diff
@@ -5,14 +5,14 @@
...
-from requests.auth import AuthBase
+from requests.auth import AuthBase, HTTPDigestAuth
 
 from . import __version__
 from .loaders import from_uri
 from .models import Case
 from .schemas import BaseSchema
 
-Auth = Union[Tuple[str, str], AuthBase]
+Auth = Union[Tuple[str, str], AuthBase, HTTPDigestAuth]
```
Stranger6667: `HTTPDigestAuth` is a subtype of `AuthBase`, it is redundant

src/schemathesis/cli/__init__.py
```diff
@@ -56,7 +64,8 @@ def main() -> None:
 @click.option("--base-url", "-b", help="Base URL address of the API.", type=str)
 def run(  # pylint: disable=too-many-arguments
     schema: str,
-    auth: Optional[Tuple[str, str]],
+    auth: Optional[Union[Tuple[str, str], AuthBase]],
```
Stranger6667: It seems like the `AuthBase` is redundant

PR #135, opened 2019-10-09
Stricter tests

Stranger6667:
Extra assertions found by mutmut + simplification of assertions re-raisung in the `runner` module


src/schemathesis/runner.py
```diff
@@ -101,10 +101,12 @@ def single_test(
             stats.increment(check_name)
         except AssertionError as e:
```
paveldedik: I'm also thinking if `AssertionError` is enough since the check can result in other errors as well (theoretically), it won't be caught in those cases. What do you think about it?

PR #150, opened 2019-10-15
Add ability to register strategies for custom string formats

### LINKED ISSUE #94: Ability to register strategies for custom string formats
E.g. we have a format `card_number` that involves Luhn algorithm validation - it would be nice to register custom hypothesis strategies for such cases. Most probably it could be solved via updating this dictionary - https://github.com/Zac-HD/hypothesis-jsonschema/blob/master/src/hypothesis_jsonschema/_impl.py#L637

On the top level it could look like this `schemathesis.register_string_format(name: str, strategy: st.SearchStrategy)`

Such customization options will improve the quality of generated test cases (however we need to pass invalid values as well, e.g. without format or pattern to verify the contrapositive assumption - invalid cases will be rejected)

svtkachenko:
Closes #94

### DIFF
--- a/src/schemathesis/_hypothesis.py
+++ b/src/schemathesis/_hypothesis.py
@@ -80,3 +80,13 @@ def get_case_strategy(endpoint: Endpoint) -> st.SearchStrategy:
         body=from_schema(endpoint.body),
         form_data=from_schema(endpoint.form_data),
     )
+
+
+def register_string_format(name: str, strategy: st.SearchStrategy) -> None:
+    if not isinstance(name, str):
+        raise TypeError(f"name must be of type {str}, not {type(name)}")
+    if not isinstance(strategy, st.SearchStrategy):
+        raise TypeError(f"strategy must be of type {st.SearchStrategy}, not {type(strategy)}")
+    from hypothesis_jsonschema._impl import STRING_FORMATS  # pylint: disable=import-outside-toplevel
+
+    STRING_FORMATS[name] = strategy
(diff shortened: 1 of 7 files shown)


src/schemathesis/_hypothesis.py
```diff
@@ -80,3 +80,9 @@ def get_case_strategy(endpoint: Endpoint) -> st.SearchStrategy:
         body=from_schema(endpoint.body),
         form_data=from_schema(endpoint.form_data),
     )
+
+
+def register_string_format(name: str, strategy: st.SearchStrategy) -> None:
```
Stranger6667: Please, add runtime validation for the arguments - we need to avoid improper usage

Stranger6667: What about string for ‘name’?

src/schemathesis/_hypothesis.py
```diff
@@ -80,3 +80,13 @@ def get_case_strategy(endpoint: Endpoint) -> st.SearchStrategy:
...
+    if not isinstance(name, str):
+        raise TypeError(f"name must be of type 'str' not {str(type(name))}")
+    if not isinstance(strategy, st.SearchStrategy):
+        raise TypeError(f"strategy must be of type {str(type(st.SearchStrategy()))}, not {str(type(strategy))}")
```
Stranger6667: There is no need to create a strategy instance to know its type:

```
>>> type(st.SearchStrategy()) == st.SearchStrategy
True
```

Stranger6667: And `str` call is redundant:

```
>>> f"type: {str(type(st.SearchStrategy()))}" == f"type: {st.SearchStrategy}"
True
```

in all f-strings in this function

Issue #119, opened 2019-10-06
Support file paths in `schema` argument in CLI


Stranger6667:
It is handy to use file paths without `file` schema:

`file:///schema.json` -> `/schema.json`

Ref: #88

PR #151, opened 2019-10-16
Use Schemathesis UserAgent during schema loading

### LINKED ISSUE #144: Use Schemathesis UserAgent during schema loading
Now it is the standard one for urllib `Python-urllib/3.7`. The whole access string for aiohttp looks like this:
`"GET /swagger.yaml HTTP/1.1" 200 616 "-" "Python-urllib/3.7"`

It should be the same user agent as in the [runner](https://github.com/kiwicom/schemathesis/blob/master/src/schemathesis/runner.py#L61) module:
`schemathesis/{__version__}` (for example `schemathesis/0.9.0`)

svtkachenko:
Closes #144

### DIFF
--- a/src/schemathesis/__init__.py
+++ b/src/schemathesis/__init__.py
@@ -1,11 +1,4 @@
-from importlib_metadata import PackageNotFoundError, version
-
 from ._hypothesis import register_string_format
+from .constants import __version__
 from .loaders import Parametrizer, from_dict, from_file, from_path, from_pytest_fixture, from_uri
 from .models import Case
-
-try:
-    __version__ = version(__package__)
-except PackageNotFoundError:
-    # Local run without installation
-    __version__ = "dev"
--- a/src/schemathesis/constants.py
+++ b/src/schemathesis/constants.py
@@ -1,0 +1,10 @@
+from importlib_metadata import PackageNotFoundError, version
+
+try:
+    __version__ = version(__package__)
+except PackageNotFoundError:
+    # Local run without installation
+    __version__ = "dev"
+
+
+USER_AGENT = f"schemathesis/{__version__}"
--- a/src/schemathesis/loaders.py
+++ b/src/schemathesis/loaders.py
@@ -1,8 +1,9 @@
 from typing import IO, Any, Dict, Optional, Union
-from urllib.request import urlopen
 
+import requests
 import yaml
 
+from .constants import USER_AGENT
 from .lazy import LazySchema
 from .schemas import BaseSchema, OpenApi30, SwaggerV20
 from .types import Filter, PathLike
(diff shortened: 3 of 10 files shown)


src/schemathesis/loaders.py
```diff
@@ -17,7 +19,7 @@ def from_path(path: PathLike, method: Optional[Filter] = None, endpoint: Optiona
 
 def from_uri(uri: str, method: Optional[Filter] = None, endpoint: Optional[Filter] = None) -> BaseSchema:
     """Load a remote resource and parse to schema instance."""
-    response = urlopen(uri)
+    response = urlopen(Request(uri, headers={"User-Agent": f"schemathesis/{schemathesis.__version__}"}))
```
Stranger6667: Can we reuse the same string in both places where ua is used? E.g some global variable (maybe in a separate ‘constants.py’ module)

Stranger6667: What do using about using requests here? In runner we can even reuse the same session if we make it as a dependency injection here

Stranger6667: E.g add an argument ‘session=None’ and if it is None, then ‘session = requests.Session()’

Stranger6667: E.g add an argument ‘session=None’ and if it is None, then ‘session = requests.Session()’

svtkachenko: urllib can be used for `file://`, while requests can't be used for that. Do we use somehow this kind of URI?

Stranger6667: Hm, no, it wasn’t documented. But ‘from_file’ loader should be used for it. Also there is an issue for file support in CLI - #119

Stranger6667: Hm, no, it wasn’t documented. But ‘from_file’ loader should be used for it. Also there is an issue for file support in CLI - #119

Stranger6667: I’d go for replacing, since it wasn’t documented and there is a proper alternative for this case

svtkachenko: Ok, on it.

src/schemathesis/__init__.py
```diff
@@ -1,11 +1,3 @@
-from importlib_metadata import PackageNotFoundError, version
-
 from ._hypothesis import register_string_format
 from .loaders import Parametrizer, from_dict, from_file, from_path, from_pytest_fixture, from_uri
 from .models import Case
-
```
Stranger6667: To be compliant with PEP-0396 we need to keep `__version__` in this module - it could be imported from `constants.py`

src/schemathesis/constants.py
```diff
@@ -0,0 +1,10 @@
+from importlib_metadata import PackageNotFoundError, version
```
Stranger6667: After moving version here the currently failing test should look like this:

```
def test_dev_version(monkeypatch, mocker):
    # When Schemathesis is run in dev environment without installation
    monkeypatch.delitem(sys.modules, "schemathesis.costants")
```

Stranger6667:
Btw, it is strange that we have 12 checks now instead of 6

Stranger6667:
Could you rebase, please?

PR #156, opened 2019-10-19
Speedup Case generation

### LINKED ISSUE #154: Speedup Case generation
Currently, we use `hypothesis.strategies.just` for static parts of `Case` like `method` or `path`. However, it will be faster to use `functools.partial` instead. Here is an example with a simple `A` class, that takes 2 arguments:

```
In [14]: %timeit st.builds(A, a=st.integers(), b=st.integers()).example()
7.87 ms ± 119 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [15]: %timeit st.builds(partial(A, 1, 1)).example()
2.43 ms ± 53.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

With more arguments, there is not that big difference when at least one strategy is passed into `builds`, but still noticeable:

```
In [21]: %timeit st.builds(A, a=st.integers(), b=st.integers(), c=st.integers()).example()
9.09 ms ± 167 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [22]: %timeit st.builds(partial(A, 1, 1), c=st.integers()).example()
7.63 ms ± 142 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

mnalband:
Closes #154

### DIFF
--- a/src/schemathesis/_hypothesis.py
+++ b/src/schemathesis/_hypothesis.py
@@ -1,5 +1,6 @@
 """Provide strategies for given endpoint(s) definition."""
 import asyncio
+from functools import partial
 from typing import Any, Callable, Generator, Optional
 
 import hypothesis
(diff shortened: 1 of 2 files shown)


Stranger6667:
However, it seems like we need to modify our GitHub Actions for forks in the way it will handle coverage correctly

svtkachenko:
@Stranger6667 Might be not possible :disappointed: 
https://github.com/codecov/codecov-action/issues/11

Stranger6667:
Ok, let’s merge this PR then

PR #157, opened 2019-10-19
Run GitHub Actions on pull request

svtkachenko:
Currently it runs only for direct push to this repository, but not for PR from forks.


Stranger6667:
Will it run on direct pushes after this change?

PR #158, opened 2019-10-19
Set default deadline to 500ms

### LINKED ISSUE #138: Set default deadline higher than 200ms
The default value for a test is 200ms. It could be not enough in some cases, e.g. auth (via argon2 for example) or proxying calls to 3rd party, probably we can increase this number by default in Schemathesis so users will have less possibility to face such cases which could be confusing at the first glance

Related: #107

mnalband:
Closes #138


Stranger6667:
Thanks! 🍰

PR #161, opened 2019-10-21
Fix CI jobs repeat on PR edit

### LINKED ISSUE #160: Run GH actions only once in PRs
Currently, they are executed twice. Example - #159

svtkachenko:
Fixes #160

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,8 +1,6 @@
 name: build
 
-on:
-  pull_request:
-    types: [opened, edited, synchronize, reopened]
+on: [pull_request]
 
 jobs:
   black:


.github/workflows/build.yml
```diff
@@ -1,8 +1,6 @@
 name: build
 
-on:
-  pull_request:
-    types: [opened, edited, synchronize, reopened]
+on: [pull_request]
```
Stranger6667: but will it work on direct pushes?

svtkachenko: No. There is no way to specify push type accroding to https://help.github.com/en/github/automating-your-workflow-with-github-actions/events-that-trigger-workflows#push-event-push

So, either we will have no CI for direct push, or no CI for PRs from forks. Or there always would be 2 runs in case of PRs: 1 for push and 1 for PR.

Stranger6667: Ok, will do PRs then :)

Issue #134, opened 2019-10-09
Filtering operations by tags


hlobit:
OpenAPI specification allow adding tags to operations.

```yaml
tags:
- pet
summary: Updates a pet in the store with form data
operationId: updatePetWithForm
parameters:
- name: petId
  in: path
  description: ID of pet that needs to be updated
  required: true
  schema:
    type: string
...
```

It would be great to filter on these tags to get a subset of cases.
Something like:

```python
@schema.parametrize(tags=['pet'])                                                 
def test_no_server_errors(case):
    # do something with the case
    pass
```

Stranger6667:
Hello @hlobit !
It is a great feature proposal :) Thank you for bringing it

Stranger6667:
Adding info for the future reference.

To implement this we need to:
- extend all relevant functions in `loaders.py` with the new `tags` argument of `Optional[Filter]` type and pass it further down;
- add a new `tags` attribute to `schemas.BaseSchema` similar to `method` & `endpoint`
- add `tags` to `lazy.LazySchema.parametrize` and `lazy.get_schema` in a similar way
- add `-t` (`tags`) option to CLI in `cli.__init__` and pass it down to runner via `loader_options`
- in `schemas.SwaggerV20.get_all_endpoints` add a check for tags on the line 105.

Not sure if it will be better to use case insensitive check from `should_skip_method` or add a separate one or reuse `should_skip_endpoint`. The latter has `re.search` type of check, which could be convenient. However, most probably it will be better to rename these functions sho they are not bound to `method` or `endpoint` but to the way they implement comparison

mnalband:
Let me take this :-)

Stranger6667:
Sure thing :)

Stranger6667:
Hej @mnalband ! Let me know if you need any support from our side :)

Stranger6667:
Done via #165

PR #165, opened 2019-10-21
Filter Swagger/OpenAPI operations by tag

mnalband:
WIP #134 Still don't know how to fix this test :-( 
==================================== ERRORS ====================================
_____________________ ERROR collecting test_tag_filter.py ______________________
/home/mnalband/schemathesis/.tox/py37/lib/python3.7/site-packages/schemathesis/extra/pytest_plugin.py:51: in collect
    item for endpoint in self.schemathesis_case.get_all_endpoints() for item in self._gen_items(endpoint)
/home/mnalband/schemathesis/.tox/py37/lib/python3.7/site-packages/schemathesis/extra/pytest_plugin.py:51: in <listcomp>
    item for endpoint in self.schemathesis_case.get_all_endpoints() for item in self._gen_items(endpoint)
/home/mnalband/schemathesis/.tox/py37/lib/python3.7/site-packages/schemathesis/schemas.py:116: in get_all_endpoints
    parameters = itertools.chain(definition.get("parameters", ()), common_parameters)
E   AttributeError: 'NoneType' object has no attribute 'get'

### DIFF
--- a/src/schemathesis/filters.py
+++ b/src/schemathesis/filters.py
@@ -1,5 +1,5 @@
 import re
-from typing import Optional
+from typing import List, Optional
 
 from .types import Filter
 from .utils import force_tuple
@@ -17,3 +17,12 @@ def should_skip_endpoint(endpoint: str, pattern: Optional[Filter]) -> bool:
         return False
     patterns = force_tuple(pattern)
     return not any(re.search(item, endpoint) for item in patterns)
+
+
+def should_skip_by_tag(tags: Optional[List[str]], pattern: Optional[Filter]) -> bool:
+    if pattern is None:
+        return False
+    if not tags:
+        return True
+    patterns = force_tuple(pattern)
+    return not any(re.search(item, tag) for item in patterns for tag in tags)
--- a/src/schemathesis/lazy.py
+++ b/src/schemathesis/lazy.py
@@ -16,17 +16,22 @@ class LazySchema:
     fixture_name: str = attr.ib()
     method: Optional[Filter] = attr.ib(default=NOT_SET)
     endpoint: Optional[Filter] = attr.ib(default=NOT_SET)
+    tag: Optional[Filter] = attr.ib(default=NOT_SET)
 
-    def parametrize(self, method: Optional[Filter] = NOT_SET, endpoint: Optional[Filter] = NOT_SET) -> Callable:
+    def parametrize(
+        self, method: Optional[Filter] = NOT_SET, endpoint: Optional[Filter] = NOT_SET, tag: Optional[Filter] = NOT_SET
+    ) -> Callable:
         if method is NOT_SET:
             method = self.method
         if endpoint is NOT_SET:
             endpoint = self.endpoint
+        if tag is NOT_SET:
+            tag = self.tag
 
         def wrapper(func: Callable) -> Callable:
             def test(request: FixtureRequest, subtests: SubTests) -> None:
                 """The actual test, which is executed by pytest."""
-                schema = get_schema(request, self.fixture_name, method, endpoint)
+                schema = get_schema(request, self.fixture_name, method, endpoint, tag)
                 fixtures = get_fixtures(func, request)
                 # Changing the node id is required for better reporting - the method and endpoint will appear there
                 node_id = subtests.item._nodeid
@@ -56,7 +61,11 @@ def run_subtest(endpoint: Endpoint, fixtures: Dict[str, Any], sub_test: Callable
 
 
 def get_schema(
-    request: FixtureRequest, name: str, method: Optional[Filter] = None, endpoint: Optional[Filter] = None
+    request: FixtureRequest,
+    name: str,
+    method: Optional[Filter] = None,
+    endpoint: Optional[Filter] = None,
+    tag: Optional[Filter] = None,
 ) -> BaseSchema:
     """Loads a schema from the fixture."""
     schema = request.getfixturevalue(name)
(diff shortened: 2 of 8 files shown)


test/test_filters.py
```diff
@@ -51,6 +51,26 @@ def test_(request, case):
...
+    testdir.make_test(
+        """
+@schema.parametrize(tag="bar")
+@settings(max_examples=1)
+def test_(request, case):
+    request.config.HYPOTHESIS_CASES += 1
+    assert case.path == "/v1/bar"
+    assert case.method == "GET"
+""",
+        paths={"/foo": {"get": parameters.update(tag="foo")}, "/bar": {"post": parameters.update(tag="bar")}},
```
Stranger6667: `dict.update` returns None and updates the dictionary in place. It should be something like this:

```
{**parameters, "tags": ["foo"]}
```

Note, there should be "tags" key according to the spec - https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-5
and it should be a list of strings.

test/test_filters.py
```diff
@@ -51,6 +51,26 @@ def test_(request, case):
     )
 
 
+def test_tag_filter(testdir):
+    # When `method` is specified
```
Stranger6667: `method` -> `tag`

test/test_filters.py
```diff
@@ -51,6 +51,26 @@ def test_(request, case):
...
+def test_(request, case):
+    request.config.HYPOTHESIS_CASES += 1
+    assert case.path == "/v1/bar"
+    assert case.method == "GET"
+""",
+        paths={"/foo": {"get": parameters.update(tag="foo")}, "/bar": {"post": parameters.update(tag="bar")}},
+    )
+    result = testdir.runpytest("-v", "-s")
+    result.assert_outcomes(passed=1)
+    # Then only tests for this method should be generated
```
Stranger6667: `method` -> `tag`

src/schemathesis/schemas.py
```diff
@@ -105,11 +109,35 @@ def get_all_endpoints(self) -> Generator[Endpoint, None, None]:
                 continue
             common_parameters = get_common_parameters(methods)
             for method, definition in methods.items():
-                if method == "parameters" or should_skip_method(method, self.method):
+                print(definition)
```
Stranger6667: Do not forget to remove the print before the merge :)

src/schemathesis/schemas.py
```diff
@@ -105,11 +109,35 @@ def get_all_endpoints(self) -> Generator[Endpoint, None, None]:
...
+                if (
+                    method == "parameters"
+                    or should_skip_method(method, self.method)
+                    or (definition and should_skip_endpoint(definition.get("tag"), self.tag))
```
Stranger6667: It seems like the first part of the condition (`definition`) will become irrelevant after fixing the schema in the failing test

Stranger6667: Also, it should be `tags` instead of `tag`

src/schemathesis/filters.py
```diff
@@ -17,3 +17,16 @@ def should_skip_endpoint(endpoint: str, pattern: Optional[Filter]) -> bool:
...
     patterns = force_tuple(pattern)
     return not any(re.search(item, endpoint) for item in patterns)
+
+
+def should_skip_by_tag(tags: Optional[List[str]], pattern: Optional[Filter]) -> bool:
+    if pattern is None:
+        return False
+    patterns = force_tuple(pattern)
+    matches = []
+    if tags:
```
Stranger6667: if there are no tags, then we know that it should be skipped - no tags to match. this check could be done just after the condition in the beginning of the function

src/schemathesis/filters.py
```diff
@@ -17,3 +17,17 @@ def should_skip_endpoint(endpoint: str, pattern: Optional[Filter]) -> bool:
...
     return not any(re.search(item, endpoint) for item in patterns)
+
+
+def should_skip_by_tag(tags: Optional[List[str]], pattern: Optional[Filter]) -> bool:
+    if pattern is None:
+        return False
+    if not tags:
+        return True
+    patterns = force_tuple(pattern)
+    matches = []
```
Stranger6667: can we have this block as a comprehension?
```
return not any(re.search(item, tag) for item in patterns for tag in tags)
```

src/schemathesis/schemas.py
```diff
@@ -105,11 +109,34 @@ def get_all_endpoints(self) -> Generator[Endpoint, None, None]:
...
+            method = self.method
+        if endpoint is NOT_SET:
+            endpoint = self.endpoint
+        if tag is NOT_SET:
+            tag = self.tag
+
+        def wrapper(func: Callable) -> Callable:
+            func._schemathesis_test = self.__class__(
+                self.raw_schema, base_url=self.base_url, method=method, endpoint=endpoint, tag=tag
+            )  # type: ignore
```
Stranger6667: This comment should be on line 133 to silence the mypy error:

```src/schemathesis/schemas.py:133: error: "Callable[..., Any]" has no attribute "_schemathesis_test"```

File: src/schemathesis/lazy.py [hunk 9]
```diff
@@ -66,6 +75,10 @@ def get_schema(
         method = schema.method
     if endpoint is NOT_SET:
         endpoint = schema.endpoint
+    if isinstance(schema, SwaggerV20):
```
Stranger6667: Now I am thinking that probably it will be easier to put `tag` field to the `BaseSchema`. To simplify everything - we can address the difference when we will actually implement support of other schema types

Stranger6667:
Great! Thank you for your contribution :)

PR #175, opened 2019-10-24
Fix error message if schema is not found in CLI

### LINKED ISSUE #172: Misleading error message if schema is not found in CLI
If schema URI returns 404, then the following error is raised and a stacktrace is shown:

```ValueError: Unsupported schema type```

But actually we don't need a stacktrace and this message is misleading - it will be better to indicate, that the schema is not found

### PR DESCRIPTION
**Pull request by svtkachenko**
Closes #172

### DIFF
--- a/test/cli/test_commands.py
+++ b/test/cli/test_commands.py
@@ -1,7 +1,9 @@
 import pytest
 from _pytest.main import ExitCode
 from hypothesis import HealthCheck, Phase, Verbosity
+from requests import Request, Response
 from requests.auth import HTTPDigestAuth
+from requests.exceptions import HTTPError
 
 from schemathesis.runner import DEFAULT_CHECKS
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: test/cli/test_commands.py [hunk 1]
```diff
@@ -237,3 +239,21 @@ def test_cli_run_output_empty(cli, schema_url):
...
+
+@pytest.mark.parametrize(
+    "status_code, message",
+    (
+        (404, f"Schema was not found via {SCHEMA_URI}"),
+        (500, f"Failed to load schema, code 500 was returned via {SCHEMA_URI}"),
+    ),
+)
+def test_execute_missing_schema(cli, mocker, status_code, message):
+    Response.status_code = status_code
```
Stranger6667: they should be instances, not classes. otherwise it could affect other tests

PR #177, opened 2019-10-24
[WIP] Run internal test suite in parallel via `pytest-xdist` by default

Stranger6667:
Also removed JUnit reports, that are not used


Stranger6667:
Will get back to it eventually, without coverage it doesn't make sense to have on CI, but it is still useful for local pytest runs

PR #178, opened 2019-10-24
Show falsifying examples in case of hypothesis errors

### LINKED ISSUE #137: Falsifying examples are not shown in case of hypothesis errors
If the test run discovers multiple errors and `hypothesis.errors.MultipleFailures: Hypothesis found 2 distinct failures.` is raised, then all the other output is missing

Also, it goes for timeout errors, we can display it as "E" in the CLI output and not stop the whole test suite

svtkachenko:
Fixes #137


src/schemathesis/runner/__init__.py
```diff
@@ -67,7 +68,11 @@ def execute_from_schema(
         for endpoint, test in schema.get_all_tests(single_test, settings):
             yield events.BeforeExecution(statistic, endpoint)
             with suppress(AssertionError):
-                test(session, base_url, checks, statistic)
+                try:
+                    test(session, base_url, checks, statistic)
+                except (hypothesis.errors.DeadlineExceeded, hypothesis.errors.MultipleFailures):
```
Stranger6667: What about other exceptions? Maybe we can put some common exception type here

docs/changelog.rst
```diff
@@ -16,6 +16,7 @@ Fixed
 
 - Test server generated wrong schema when ``endpoints`` option is passed via CLI. `#173`_
 - Error message if schema is not found in CLI. `#172`_
+- Show falsifying examples in case of hypothesis errors `#137`_
```
Stranger6667: Please, rephrase. It should be about handling Hypothesis errors. Falsifying examples are not that relevant, the important part is that the test suite doesn't stop on such errors

Stranger6667:
Please, fix the coverage as well

PR #179, opened 2019-10-25
Update CI jobs and add scheduled workflow with mutmut

svtkachenko:


Stranger6667:
I think we need to add `# pragma: no mutate` comments in the appropriate places to reduce the total mutations number

tox.ini
```diff
@@ -29,3 +29,9 @@ commands =
     coverage combine
     coverage report
     coverage xml -i {posargs:}
+
+[testenv:mutmut]
+basepython = python3.7
+deps = mutmut
+skip_install = true
+commands = mutmut run --backup --paths-to-mutate "src/" --tests-dir "test/" --runner "pytest -x"
```
Stranger6667: We can also add `--tb=no -q` to the pytest command to speed up it a little

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 isolated_build = true
-envlist = pylint,mypy,py{36,37,38},coverage-report
+envlist = pylint,mypy,py{36,37,38},coverage-report,mutmut
```
Stranger6667: Not sure if it is the right place to add a command, I am concerning regarding running `tox -p all -o` - it will include the `mutmut` build as well, it could be manually excluded, but I think it shouldn't be run by default at all. What do you think?

PR #182, opened 2019-10-26
Fix percentage output & make CLI output closer to pytest style

Stranger6667:
TODO:

- [ ] Display exceptions from Hypothesis. Now, e.g. deadline missing is not shown, but only falsifying example for this case. Maybe there should be two sections - "hypothesis report" and "hypothesis errors"


Stranger6667:
Will complete the TODO part in a separate PR

PR #205, opened 2019-10-31
Option to suppress schema errors

### LINKED ISSUE #139: Option to suppress schema errors
There are 2 cases:
- The schema contains an error that is unveiled during strategy generation  (e.g. "int" type instead of "integer") 
- The schema contains an error that is unveiled during the test execution (e.g. missing "required=true" in parameters with "in=path")

If might be useful to have an ability to continue on such errors, or maybe even have this behavior as default with an option to stop on first error.

Examples of both cases - https://httpbin.org/spec.json

svtkachenko:
Resolves #139

### DIFF
--- a/src/schemathesis/extra/pytest_plugin.py
+++ b/src/schemathesis/extra/pytest_plugin.py
@@ -41,7 +41,7 @@ class SchemathesisCase(PyCollector):
         try:
             hypothesis_item = create_test(endpoint, self.test_function)
         except InvalidEndpoint:
-            pytest.fail("Invalid endpoint")
+            pytest.fail("Invalid schema for this endpoint")
         items = self.ihook.pytest_pycollect_makeitem(
             collector=self.parent, name=self._get_test_name(endpoint), obj=hypothesis_item
         )
--- a/src/schemathesis/extra/pytest_plugin.py
+++ b/src/schemathesis/extra/pytest_plugin.py
@@ -6,6 +6,8 @@ from _pytest.config import hookimpl
 from _pytest.python import Function, PyCollector  # type: ignore
 from hypothesis.errors import InvalidArgument  # pylint: disable=ungrouped-imports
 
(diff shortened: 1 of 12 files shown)


src/schemathesis/extra/pytest_plugin.py
```diff
@@ -36,7 +38,10 @@ def _gen_items(self, endpoint: Endpoint) -> Generator[Function, None, None]:
         Could produce more than one test item if
         parametrization is applied via ``pytest.mark.parametrize`` or ``pytest_generate_tests``.
         """
-        hypothesis_item = create_test(endpoint, self.test_function)
+        try:
+            hypothesis_item = create_test(endpoint, self.test_function)
+        except InvalidEndpoint:
+            pytest.fail("Invalid endpoint")
```
Stranger6667: It would be nice to have some more verbose error definition, e.g. that there is an error in the schema + maybe some info about from the underlaying exception

src/schemathesis/runner/__init__.py
```diff
@@ -66,6 +66,8 @@ def execute_from_schema(
         yield events.Initialized(results=results, schema=schema, checks=checks, hypothesis_settings=settings)
 
         for endpoint, test in schema.get_all_tests(single_test, settings):
+            if not endpoint.is_valid:
```
Stranger6667: In this case, we should not run the test later on - we can put this condition to `try/except`:

```
try:
    if endpoint.is_valid:
        test(session, base_url, checks, result)
        status = Status.success
    else:
        status = Status.invalid
except AssertionErrpr:
    ...
```

src/schemathesis/schemas.py
```diff
@@ -67,7 +69,12 @@ def get_all_tests(
     ) -> Generator[Tuple[Endpoint, Callable], None, None]:
```
Stranger6667: It should be `Optional[Callable]`

PR #234, opened 2019-11-04
Generate binary data for `format=binary`

### LINKED ISSUE #197: Generate binary data for `format=binary`
It might be useful for testing of uploading files.

To do so we need to define & register a custom strategy via `register_string_format`, something like this:

```
st.builds(io.BytesIO, st.binary())
```

svtkachenko:
Closes #197

### DIFF
--- a/src/schemathesis/__init__.py
+++ b/src/schemathesis/__init__.py
@@ -1,4 +1,6 @@
-from ._hypothesis import register_string_format
+from ._hypothesis import init_default_strategies, register_string_format
 from .constants import __version__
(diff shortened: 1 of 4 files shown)


src/schemathesis/__init__.py
```diff
@@ -1,4 +1,6 @@
-from ._hypothesis import register_string_format
+from ._hypothesis import init_default_strategis, register_string_format
 from .constants import __version__
 from .loaders import Parametrizer, from_dict, from_file, from_path, from_pytest_fixture, from_uri
 from .models import Case
+
+init_default_strategis()
```
Stranger6667: typo - strategies

PR #252, opened 2019-11-11
Support OpenAPI 3.0 base paths

### LINKED ISSUE #40: Check alternative base paths
In OAS 3 it could be specified per server:

```
servers:
  - url: http://localhost:8080/{basePath}
    description: application
    variables:
      basePath:
        default: api
```

Currently, Schemathesis only uses `basePath` from the schema root, which could not work on some OAS schemas

If we will automate this, then it will reduce the number of actions needed to run the tests, currently, the base URL should be set manually in a test case

svtkachenko:
Closes #40

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,11 @@ Changelog
 `Unreleased`_
 -------------
 
+Added
+~~~~~
+
+- Support for OpenAPI 3.0 server variables (base_path) `#40`_
+
 `0.14.0`_ - 2019-11-09
 ----------------------
 
@@ -434,6 +439,7 @@ Fixed
 .. _#58: https://github.com/kiwicom/schemathesis/issues/58
 .. _#55: https://github.com/kiwicom/schemathesis/issues/55
 .. _#45: https://github.com/kiwicom/schemathesis/issues/45
+.. _#40: https://github.com/kiwicom/schemathesis/issues/40
 .. _#35: https://github.com/kiwicom/schemathesis/issues/35
 .. _#34: https://github.com/kiwicom/schemathesis/issues/34
 .. _#30: https://github.com/kiwicom/schemathesis/issues/30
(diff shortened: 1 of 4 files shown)


src/schemathesis/schemas.py
```diff
@@ -263,6 +263,22 @@ def spec_version(self) -> str:
...
+    def base_path(self) -> str:
+        """Base path for the schema."""
+        servers = self.raw_schema.get("servers", [])
+        try:
+            # assume we're the first server in list
+            server = deepcopy(servers[0])
+            server_vars = server.pop("variables", {})
+            server["url"] = server["url"].format(**{k: v["default"] for k, v in server_vars.items()})
+            path = urlsplit(server["url"]).path
+        except IndexError:
```
Stranger6667: I think it will be much more clear to have explicit checks here

docs/changelog.rst
```diff
@@ -11,6 +11,7 @@ Added
 
 - CLI: Support file paths in ``schema`` argument. `#119`_
 - Check to verify response status & content type in CLI / Runner. `#101`_
+- Support for OpenAPI 3.0 server variables (base_path) `#40`_
```
Stranger6667: this version is already released, please, put it into another section

src/schemathesis/schemas.py
```diff
@@ -263,6 +263,22 @@ def spec_version(self) -> str:
...
 
+    @property
+    def base_path(self) -> str:
+        """Base path for the schema."""
+        servers = self.raw_schema.get("servers", [])
+        if servers:
+            # assume we're the first server in list
+            server = deepcopy(servers[0])
+            server_vars = server.pop("variables", {})
+            server["url"] = server["url"].format(**{k: v["default"] for k, v in server_vars.items()})
```
Stranger6667: do we need this assignment in the dict? it seems like `deepcopy` is not needed as well

PR #268, opened 2019-11-14
Create a docker image for CLI

svtkachenko:
Closes #268

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+**/.dockerignore
+**/.env
+**/.git
+**/.gitignore
+**/docker-compose*
+**/Dockerfile*
+**/docs
+**/img
+**/test
+LICENSE
--- a/.github/workflows/master_update.yml
+++ b/.github/workflows/master_update.yml
@@ -1,0 +1,18 @@
+name: Perform actions on master branch update
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release_image:
+    name: Build and publish to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: elgohr/Publish-Docker-Github-Action@master
+      with:
+        name: kiwicom/schemathesis:latest
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_PASSWORD }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,17 @@
+name: Perform actions after release
+
+on:
+  release:
+    type: [published, created, edited]
+
+jobs:
+  release_image:
+    name: Build and publish to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: elgohr/Publish-Docker-Github-Action@master
+      with:
+        name: kiwicom/schemathesis:$GITHUB_REF
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_PASSWORD }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM python:3.8-alpine
+
+LABEL Name=Schemathesis
+
+RUN python3 -m pip install --upgrade pip && \
+    python3 -m pip --no-cache-dir install poetry
+
+WORKDIR /app
+RUN addgroup -S schemathesis && \
+    adduser -D -S schemathesis -G schemathesis -s /sbin/nologin && \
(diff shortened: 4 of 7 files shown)


.github/workflows/master_update.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  release_image:
+    name: Build and publish to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: elgohr/Publish-Docker-Github-Action@master
+      with:
+        name: svtkachenko/schemathesis:latest
```
Stranger6667: hm, should it be `svtkachenko` or `kiwicom`?

svtkachenko: It is for a purpose of testing, see `[WIP]` in header :slightly_smiling_face:

.dockerignore
```diff
@@ -0,0 +1,26 @@
+**/.classpath
+**/.dockerignore
+**/.env
+**/.git
+**/.gitignore
+**/.project
+**/.settings
+**/.toolstarget
+**/.vs
+**/.vscode
```
aexvir: While true that usually those should be excluded, in the context of CI you don't usually have any of the files excluded using the `.gitignore`, so I would probably not duplicate them.

That can cause bigger images when building it locally, but usually you don't want to do that either way, except for testing/debugging purposes

.dockerignore
```diff
@@ -0,0 +1,26 @@
...
+**/*.*proj.user
+**/*.dbmdl
+**/*.jfm
+**/azds.yaml
+**/bin
+**/charts
+**/docker-compose*
+**/Dockerfile*
+**/node_modules
```
aexvir: 🤔 I'm not seeing any `package.json`, is this folder usually in the context? Or the `.dockerignore` was generated with some tool?

.github/workflows/master_update.yml
```diff
@@ -0,0 +1,18 @@
...
+    # branches:
+      # - master
+
+jobs:
+  release_image:
+    name: Build and publish to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: elgohr/Publish-Docker-Github-Action@master
```
aexvir: I never used a GitHub action for this, personally I prefer to make my own job pushing to CI, but this one doesn't look bad. I'd also take advantage of [caching](https://github.com/marketplace/actions/publish-docker#cache), to avoid unnecessary slowdowns.

So... no need to change anything here, LGTM, I just wanted to let you know about this 🙂

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.8-alpine
+
+LABEL Name=Schemathesis
+
+WORKDIR /app
+ADD . /app
```
aexvir: It's usually better practice to use `COPY` instead of `ADD`, as `ADD` does fancy stuff, like resolve URLs, zips, and what not. It can have unexpected side effects

.dockerignore
svtkachenko: Yeas, it was autogenerated by VS Code.

Dockerfile
aexvir: I'd also recommend using relative paths to the `WORKDIR`, as it's easier to just change the `WORKDIR` and have all the paths adjusting automatically instead of having to update the same string in multiple places 🙂

.github/workflows/release.yml
```diff
@@ -0,0 +1,17 @@
...
+on:
+  release:
+    type: [published, created, edited]
+
+jobs:
+  release_image:
+    name: Build and publish to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
```
Stranger6667: Should it be master? because I can make a tag, then add few commits to master and then make a release from that old tag

svtkachenko: 1. In this line `master` is reference to action repo branch, i.e. `checkout master branch from https://github.com/actions/checkout`, not `checkout master from https://github.com/kiwicom/schemathesis/`.
Maybe it is better to use release tags, e.g. `actions/checkout@v1` but this is the same as in `build.yml`

To specify specific branch/tag for this repo syntax would be:
```
- uses: actions/checkout@v1
  with:
    ref: some-branch
```

More here: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#example-using-versioned-actions

2. By default checkout action checkouts specific commit, which hash is stored in `GITHUB_SHA` and it depends on event that triggered job. Here it is triggered by release, release always is associated with tag and in this case `GITHUB_SHA` is last commit in the tagged release.

More info: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#release-event-release

PR #273, opened 2019-11-15
Display seed in the CLI output.

### LINKED ISSUE #267: Display seed in the CLI output
It will help to simplify errors reproducing

svtkachenko:
Closes #267 

Output example:
```
=== ERRORS ===
____ GET: /api/flaky ____
hypothesis.errors.Flaky: Tests on this endpoint produce unreliable results: 
Falsified on the first call but did not on a subsequent one

Used seed       : 246442565675977375928223543448485731889
Query           : {'id': 0}

Run this Python code to reproduce this failure: 

requests.get('http://127.0.0.1:8081/api/flaky', params={'id': 0})
=== FAILURES ===
___ GET: /api/failure ____
Received a response with 5xx status code: 500

Used seed       : 313022266096038506880814748301664910156
Check           : not_a_server_error

Run this Python code to reproduce this failure: 

requests.get('http://127.0.0.1:8081/api/failure')
```

### DIFF
--- a/src/schemathesis/models.py
+++ b/src/schemathesis/models.py
@@ -147,6 +147,7 @@ class TestResult:
     checks: List[Check] = attr.ib(factory=list)  # pragma: no mutate
     errors: List[Tuple[Exception, Optional[Case]]] = attr.ib(factory=list)  # pragma: no mutate
     is_errored: bool = attr.ib(default=False)  # pragma: no mutate
+    seed: Optional[int] = attr.ib(default=None)  # pragma: no mutate
 
     def mark_errored(self) -> None:
         self.is_errored = True
--- a/src/schemathesis/runner/__init__.py
+++ b/src/schemathesis/runner/__init__.py
@@ -115,6 +115,7 @@ def execute_from_schema(
     auth: Optional[Auth] = None,
     headers: Optional[Dict[str, Any]] = None,
     request_timeout: Optional[int] = None,
+    seed: Optional[int] = None,
 ) -> Generator[events.ExecutionEvent, None, None]:
     """Execute tests for the given schema.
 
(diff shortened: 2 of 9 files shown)


src/schemathesis/models.py
```diff
@@ -147,6 +147,7 @@ class TestResult:
     checks: List[Check] = attr.ib(factory=list)  # pragma: no mutate
     errors: List[Tuple[Exception, Optional[Case]]] = attr.ib(factory=list)  # pragma: no mutate
     is_errored: bool = attr.ib(default=False)  # pragma: no mutate
+    seed: Optional[int] = attr.ib(default=None)  # pragma: no mutate
```
Stranger6667: I am not sure if it should be optional since there is always some seed in hypothesis

svtkachenko: If there is a hypothesis test. In case of invalid schema there is no test.

Stranger6667: aha, got it

src/schemathesis/runner/__init__.py
```diff
@@ -164,6 +164,7 @@ def execute_from_schema(
             except Exception as error:
                 status = Status.error
                 result.add_error(error)
+            result.seed = getattr(test, "_hypothesis_internal_use_generated_seed", None)
```
Stranger6667: It will be better to set it when `result` is created

svtkachenko: `test._hypothesis_internal_use_generated_seed` does not exist at that moment. Hypothesis creates is during test execution

Stranger6667: Got it, could you add a comment here?

PR #281, opened 2019-11-17
Fix pytest fixtures parametrization via `pytest_generate_tests` hook

### LINKED ISSUE #280: Pytest fixture parametrization doesnt work properly
Example:

```
def pytest_generate_tests(metafunc):
    metafunc.parametrize("inner", ("A", "B"))

@pytest.fixture()
def param(inner):
    return inner * 2

@schema.parametrize()
def test_api(request, param, case):
    ...
```

With the current implementation test run will fail with ```In test_: function uses no argument 'inner'```

However, it should properly parametrize the fixture. A real-world example is `pytest_generate_tests` from aiohttp pytest plugin - https://github.com/aio-libs/aiohttp/blob/master/aiohttp/pytest_plugin.py#L174

Stranger6667:
Fixes #280

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -383,7 +383,8 @@ Fixed
 
 - Initial public release
 
-.. _Unreleased: https://github.com/kiwicom/schemathesis/compare/v0.14.0...HEAD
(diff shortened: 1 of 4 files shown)


Stranger6667:
Unfortunately, the new uncovered line can't be properly tested without fixing #282. I am going to address it in a separate PR

PR #283, opened 2019-11-18
[WIP] Support for in-process WSGI apps testing

### LINKED ISSUE #31: Test case executor for WSGI apps
Similar to what Flask has for testing - we don't have to use `requests`, we can work with the application directly. For this, we need to make this component interchangeable and write a one that will be able to execute schemathesis tests against an arbitrary WSGI app.

In general, this approach should be much faster than using a real network. The executor API design is up to discussion (feel free to propose any).

The main point of the issue is to simplify Schemathesis adoption - it is easier and faster to use a WSGI instance in certain frameworks like Flask or Django.

The implementation could be adopted from existing test clients, for example from Flask.

https://github.com/pallets/flask/blob/master/src/flask/testing.py#L115

Or, maybe we can use `werkzeug` directly - https://github.com/pallets/werkzeug/blob/master/src/werkzeug/test.py#L768

**PEP**: https://www.python.org/dev/peps/pep-3333/

svtkachenko:
Closes #31 

TODO:
- [x] Add more test cases for `validate_app`
- [x] Fix coverage
- [x] Test with multiple workers (probably it could be extra parametrization level)
- [x] Make flask server behave in the same way as aiohttp one (the same port, args, etc). add CLI option to `test_server.sh`
- [x] catch output from the application - it should not mess the CLI, not it just appears among other symbols. should be in a separate section
- [x] fix mypy on CI
- [x] Do not display `App` in the CLI error output
- [x] Display proper reproduction code - it should be `werkzeug`, not `requests` - POSTPONED
- [x] refactor runner - code for WSGI apps should not create `requests.Session`
- [x] handle Auth & headers in runner properly - they should be propagated to the `werkzeug` client
- [x] checks should receive `Union[requests.Response, WSGIResponse]` or some another class that will provide consistent behavior + verify that all check works
- [x] Handle 404 in `from_wsgi`
- [x] Docs


pyproject.toml
```diff
@@ -44,6 +44,8 @@ pytest-mock = "^1.11.0"
 pytest-asyncio = "^0.10.0"
 pytest-xdist = "^1.30"
 typing_extensions = "^3.7"
+flask = "^1.1"
+werkzeug = "^0.16.0"
```
Stranger6667: Most probably we will need `werkzeug` in common dependencies, not only in dev ones

src/schemathesis/loaders.py
```diff
@@ -83,6 +84,13 @@ def from_pytest_fixture(
     return LazySchema(fixture_name, method=method, endpoint=endpoint, tag=tag)
 
 
+def from_wsgi(app, schema_path):
+    from werkzeug import Client  # pylint: disable=import-outside-toplevel
```
Stranger6667: it can be on the top level

src/schemathesis/loaders.py
```diff
@@ -83,6 +84,13 @@ def from_pytest_fixture(
...
+
+    client = Client(app, BaseResponse)
+    return from_file(client.get(schema_path).data)
```
Stranger6667: could you extract response in a separate variable, please?

```
response = client.get(schema_path)
```

src/schemathesis/loaders.py
```diff
@@ -83,6 +85,12 @@ def from_pytest_fixture(
     return LazySchema(fixture_name, method=method, endpoint=endpoint, tag=tag)
 
 
+def from_wsgi(app, schema_path):
+    client = Client(app, BaseResponse)
+    response = client.get(schema_path)
+    return from_file(response.data)
```
Stranger6667: I propose to pass `client` to all other loaders and store it as an optional attribute on `BaseSchema`. We can pass it to `Case` instances later and then use it for calling the app via `case.wsgi_call`

src/schemathesis/models.py
```diff
@@ -87,6 +90,19 @@ def as_requests_kwargs(self, base_url: Optional[str] = None) -> Dict[str, Any]:
...
+        extra: Dict[str, Optional[Dict]]
+        if self.form_data:
+            extra = {"data": self.form_data}
+        else:
+            extra = {"json": self.body}
+        return {
+            "method": self.method,
+            "path": self.formatted_path,
+            "headers": self.headers,
+            "query_string": self.query ** extra,
```
Stranger6667: seems like a missing `,` before `**`

src/schemathesis/schemas.py
```diff
@@ -82,6 +84,7 @@ def get_all_tests(
         """Generate all endpoints and Hypothesis tests for them."""
         test: Union[Callable, InvalidSchema]
         for endpoint in self.get_all_endpoints():
+            endpoint.wsgi_client = self.wsgi_client
```
Stranger6667: I think it should be where the `endpoint` is actually created, in `make_endpoint`

Stranger6667:
TODO:

- [ ] Refactor `Case` class (maybe add `client` entity and delegate some logic there)
- [ ] Test common execution scenarios - e.g. form data, cookies, etc
- [ ] Add support to the runner. And verify if everything works with multiple workers
- [ ] Add option to CLI
- [ ] Extend testing app - use Flask version (+ maybe Django version). It should be available as a CLI option
- [ ] Test against Django application as well

README.rst
```diff
@@ -236,6 +238,31 @@ To narrow down the scope of the schemathesis tests it is possible to filter by m
...
+
+    @app.route("/v1/users", methods=["GET"])
+    def users():
+        return jsonify([{"name": "Robin"}])
+
+    schema = schemathesis.from_wsgi("/schema.json", app)
+
+    @schema.parametrize()
+    def test_no_server_errors(case):
+        response = case.call_wsgi()
```
Stranger6667: It will be nice to support an explicit app as an argument to call_wsgi

src/schemathesis/cli/__init__.py
```diff
@@ -168,11 +171,12 @@ def run(  # pylint: disable=too-many-arguments
...
         if pathlib.Path(schema).is_file():
-            prepared_runner = runner.prepare(
-                schema, checks=selected_checks, workers_num=workers_num, loader=from_path, **options
-            )
+            options.update({"checks": selected_checks, "workers_num": workers_num, "loader": from_path})
         else:
-            prepared_runner = runner.prepare(schema, checks=selected_checks, workers_num=workers_num, **options)
+            if app is not None:
+                options["loader"] = from_wsgi
+            options.update({"checks": selected_checks, "workers_num": workers_num})
```
Stranger6667: Seems like checks and workers_num are present in both branches - could be moved before the condition

src/schemathesis/cli/callbacks.py
```diff
@@ -1,14 +1,18 @@
...
 
 import click
 import hypothesis
 
+from schemathesis import utils
+
 
 def validate_schema(ctx: click.core.Context, param: click.core.Parameter, raw_value: str) -> str:
-    if not urlparse(raw_value).netloc:
+    if "app" not in ctx.params and not urlparse(raw_value).netloc:
```
Stranger6667: How the schema path should be validated if app is present? Could it be a usual validation? Or somehow adjusted - e.g we can allow full urls to load them via network and paths to load from the app itself

src/schemathesis/cli/callbacks.py
```diff
@@ -30,6 +34,25 @@ def validate_base_url(ctx: click.core.Context, param: click.core.Parameter, raw_
...
+    path, name = (re.split(r":(?![\\/])", raw_value, 1) + [None])[:2]  # type: ignore
+    try:
+        __import__(path)
+    except (ImportError, ValueError):
+        raise click.BadParameter("Can not import application from the given module")
+    except Exception as exc:
+        message = utils.format_exception(exc)
+        click.secho(f"Error: {message}", fg="red")
+        raise click.Abort
+    module = sys.modules[path]
```
Stranger6667: Maybe we can reuse the result from import and avoid using sys

src/schemathesis/models.py
```diff
@@ -21,13 +25,14 @@ class Case:
...
-    headers: Headers = attr.ib(default=None)  # pragma: no mutate
-    cookies: Cookies = attr.ib(default=None)  # pragma: no mutate
-    query: Query = attr.ib(default=None)  # pragma: no mutate
+    path_parameters: Optional[PathParameters] = attr.ib(default=None)  # pragma: no mutate
+    headers: Optional[Headers] = attr.ib(default=None)  # pragma: no mutate
+    cookies: Optional[Cookies] = attr.ib(default=None)  # pragma: no mutate
+    query: Optional[Query] = attr.ib(default=None)  # pragma: no mutate
     body: Optional[Body] = attr.ib(default=None)  # pragma: no mutate
-    form_data: FormData = attr.ib(default=None)  # pragma: no mutate
+    form_data: Optional[FormData] = attr.ib(default=None)  # pragma: no mutate
```
Stranger6667: These changes could be done outside of this PR

src/schemathesis/models.py
```diff
@@ -37,9 +42,20 @@ def formatted_path(self) -> str:
...
+                raise ValueError(
+                    "Base URL is required as `base_url` argument in `call` or should be specified "
+                    "in the schema constructor as a part of Schema URL."
+                )
+        return base_url
+
     def get_code_to_reproduce(self) -> str:
-        """Construct a Python code to reproduce this case with `requests`."""
-        kwargs = self.as_requests_kwargs()
+        base_url = self.base_url or "http://localhost"
```
Stranger6667: It should be a proper code for werkzeug and not requests one with localhost

src/schemathesis/runner/__init__.py
```diff
@@ -310,17 +435,43 @@ def prepare(  # pylint: disable=too-many-arguments
...
+    case: Case,
+    checks: Iterable[Callable],
+    result: TestResult,
+    auth: Optional[Tuple[str, str]],
+    auth_type: Optional[str],
+    headers: Dict[str, Any],
+) -> None:
+    # pylint: disable=too-many-arguments
+    headers = headers or {}
+    headers.setdefault("User-agent", USER_AGENT)
```
Stranger6667: Headers preparation could be extracted into a separate function

src/schemathesis/runner/checks.py
```diff
@@ -1,31 +1,35 @@
...
+from typing import Callable, Tuple, Union
 
 import jsonschema
 import requests
 
 from ..models import TestResult
-from ..utils import are_content_types_equal
+from ..utils import WSGIResponse, are_content_types_equal
 
+GenericResponse = Union[requests.Response, WSGIResponse]
```
Stranger6667: Shouldn’t be mutated - add a comment for mutmut

src/schemathesis/runner/checks.py
```diff
@@ -1,31 +1,35 @@
...
 
-def not_a_server_error(response: requests.Response, result: TestResult) -> None:
+
+def not_a_server_error(response: GenericResponse, result: TestResult) -> None:
     """A check to verify that the response is not a server-side error."""
-    assert response.status_code < 500, f"Received a response with 5xx status code: {response.status_code}"
+    if response.status_code >= 500:
+        raise AssertionError(f"Received a response with 5xx status code: {response.status_code}")
```
Stranger6667: This one could be done separately - quite strange behavior. Asserts here were rewritten by pytest for some reason

src/schemathesis/runner/checks.py
```diff
@@ -60,8 +64,12 @@ def response_schema_conformance(response: requests.Response, result: TestResult)
     schema = definition.get("schema")
     if not schema:
         return
+    if isinstance(response, requests.Response):
```
Stranger6667: Maybe it will be better to provide some kind of abstraction so code in checks will be unified

test/apps/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+@click.option("--framework", type=click.Choice(["aiohttp", "flask"]), default="aiohttp")
+def run_app(port: int, endpoints: List[Endpoint], framework: str) -> None:
+    if endpoints is not None:
+        prepared_endpoints = tuple(endpoint.name for endpoint in endpoints)
+    else:
+        prepared_endpoints = tuple(endpoint.name for endpoint in Endpoint)
+    if framework == "aiohttp":
+        app = _aiohttp.create_app(prepared_endpoints)
+        web.run_app(app, port=port)
+    elif framework == "flask":
```
Stranger6667: It will be nice to test Django as well - maybe in a separate PR

test/test_models.py
```diff
@@ -11,7 +11,7 @@ def test_path():
 
 @pytest.mark.parametrize("override", (False, True))
 @pytest.mark.parametrize("converter", (lambda x: x, lambda x: x + "/"))
-def test_as_requests_kwargs(override, server, base_url, converter):
+def test_as_kwargs(override, server, base_url, converter):
```
Stranger6667: The previous name could be still used

src/schemathesis/cli/callbacks.py
```diff
@@ -1,14 +1,17 @@
...
+import re
 from contextlib import contextmanager
-from typing import Dict, Generator, Optional, Tuple
+from typing import Any, Dict, Generator, Optional, Tuple
 from urllib.parse import urlparse
 
 import click
 import hypothesis
 
+from schemathesis import utils
```
Stranger6667: Relative import

src/schemathesis/models.py
```diff
@@ -37,9 +42,20 @@ def formatted_path(self) -> str:
...
+                base_url = self.base_url
+            else:
+                raise ValueError(
+                    "Base URL is required as `base_url` argument in `call` or should be specified "
+                    "in the schema constructor as a part of Schema URL."
+                )
+        return base_url
+
     def get_code_to_reproduce(self) -> str:
-        """Construct a Python code to reproduce this case with `requests`."""
```
Stranger6667: Docstring could be restored

src/schemathesis/models.py
```diff
@@ -37,9 +42,20 @@ def formatted_path(self) -> str:
         except KeyError:
             raise InvalidSchema("Missing required property `required: true`")
 
+    def _get_base_url(self, base_url: Optional[str] = None) -> str:
```
Stranger6667: The move is not necessary

src/schemathesis/models.py
```diff
@@ -90,20 +95,65 @@ def as_requests_kwargs(self, base_url: Optional[str] = None) -> Dict[str, Any]:
     def call(
         self, base_url: Optional[str] = None, session: Optional[requests.Session] = None, **kwargs: Any
     ) -> requests.Response:
-        """Convert the case into a dictionary acceptable by requests."""
```
Stranger6667: Docstring could be restored

src/schemathesis/models.py
```diff
@@ -90,20 +95,65 @@ def as_requests_kwargs(self, base_url: Optional[str] = None) -> Dict[str, Any]:
...
         if session is None:
             session = requests.Session()
             close_session = True
         else:
             close_session = False
 
         base_url = self._get_base_url(base_url)
-        data = self.as_requests_kwargs(base_url)
```
Stranger6667: not necessary

PR #285, opened 2019-11-18
Refactor case generation strategies

### LINKED ISSUE #253: Refactor case generation strategies
At the moment we use a schema for an empty object - `{"properties": {}, "additionalProperties": False, "type": "object", "required": []}` for all `Endpoint` attributes that we generate afterwards, also, `body` is replaced by actual object, which is taken from the schema.

The problem is that it might ambiguous to differentiate between the absence of value in the generated `case` or an empty value. For example, `body` could be `None` or `{}` or `[]` which all are falsy values in Python - this should be unified and only a single value should indicate the absence. It causes problems in the CLI output, now there is hardcoded `{}`  in `display_example`, and thus `[]` sent in the body will not be displayed. 

Also, the current approach implies the usage of `from_schema`, however, we know upfront that there will be no value but Hypothesis generates data for this specification which might take time - calling `.example()` on versions with empty objects generation & without it take `2.65ms` and `1.15ms` respectively, which is a visible difference. We can optimize this by not using strategies for such cases at all - we already know what should be the value there.

svtkachenko:
Closes #253

### DIFF
--- a/src/schemathesis/_hypothesis.py
+++ b/src/schemathesis/_hypothesis.py
@@ -14,7 +14,7 @@ from requests.utils import check_header_validity  # type: ignore
 
 from ._compat import handle_warnings
 from .exceptions import InvalidSchema
-from .models import Case, Endpoint, empty_object
+from .models import Case, Endpoint
 
 PARAMETERS = frozenset(("path_parameters", "headers", "cookies", "query", "body", "form_data"))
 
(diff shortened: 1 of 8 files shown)


src/schemathesis/_hypothesis.py
```diff
@@ -61,7 +61,7 @@ def async_run(*args: Any, **kwargs: Any) -> None:
 def get_examples(endpoint: Endpoint) -> Generator[Case, None, None]:
     for name in PARAMETERS:
         parameter = getattr(endpoint, name)
-        if "example" in parameter:
+        if parameter and "example" in parameter:
```
Stranger6667: could it be `if parameter is not None and "example" in parameter`?

src/schemathesis/_hypothesis.py
```diff
@@ -114,17 +118,23 @@ def get_case_strategy(endpoint: Endpoint) -> st.SearchStrategy:
...
+                if parameter == "path_parameters":
+                    strategies[parameter] = (
+                        from_schema(endpoint.path_parameters)
+                        .filter(filter_path_parameters)  # type: ignore
+                        .map(quote_all)  # type: ignore
+                    )
+                elif parameter == "headers":
+                    strategies[parameter] = from_schema(endpoint.headers).filter(is_valid_header)  # type: ignore
+                else:
+                    strategies[parameter] = from_schema(getattr(endpoint, parameter))  # type: ignore
```
Stranger6667: could you, please define `value = getattr(endpoint, parameter)` and use it in all branches ?

PR #293, opened 2019-11-20
build(docker): Make image more usable in CI environments

aexvir:
* Improve caching by just copying essential files for each stage
* Group RUN commands in as few layers as possible
* Use `poetry==1.0.0b5`, which allows avoiding the installation of the local package
* Generate `schemathesis` comand, globally accessible
* Use whitelist approach for `.dockerignore`

Apologies for not noticing that it wouldn't be easy to run it on CI during my first review

I don't fully like the solution, but unfortunately `poetry` is not yet ready for using it comfortably outside of the dev environment. Personally, when I've used poetry in the past I just ended up generating a `requirements.txt` and using that in the docker image instead... but with this changes it's enough to make it usable both in CI and locally

locally, nothing changed:
```bash
docker run --rm kiwicom/schemathesis run http://localhost/openapi.json
```

In (gitlab) CI:
```yaml
schemathesis
  image:
    name: kiwicom/schemathesis
    entrypoint: [""]
  stage: test
  services:
    - name: $CI_REGISTRY_IMAGE:$CI_COMMIT_SHA
      alias: backend
  script:
    - schemathesis run --base-url=http://backend$ENDPOINT $SCHEMA
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,10 +1,4 @@
-**/.dockerignore
-**/.env
-**/.git
-**/.gitignore
-**/docker-compose*
-**/Dockerfile*
-**/docs
-**/img
-**/test
-LICENSE
+*
+!poetry.lock
+!pyproject.toml
+!src/schemathesis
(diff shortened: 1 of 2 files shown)


Stranger6667:
Thank you!

PR #300, opened 2019-11-23
Concurrent test execution in CLI / runner

### LINKED ISSUE #91: Runner. Execute tests for different endpoints in parallel
Now, `runner` module executes tests sequentially, but for each method/endpoint combo they could be executed in parallel. However, apparently it could be done in separate processes since `hypothesis` has some internal state that prevents running it in different threads / same thread (via `asyncio`). The network part could be done async, but not sure how to design it in this way - I did two attempts but failed because off hitting `deadline` in `hypothesis` (maybe disabling the deadline setting will help, not sure) and didn't properly try different processes. Also, it would be nice to consult with `hypothesis` core devs

Stranger6667:
Resolves #91 

It is a very rough draft

TODO:

- [x] Display number of workers in the CLI output
- [x] Add option to the CLI
- [x] Add different output styles. For concurrent execution, we can do the same as `pytest-xdist` and display only resulting symbols (e.g. `F`, or `.`)
- [x] Handle main thread interruptions
- [x] Handle interruptions in non-main threads
- [x] Check how not handled exceptions will be displayed in worker threads;
- [x] Move Hypothesis output that happens in worker threads to the main thread;
- [x] Test things 
- [x] Docs

Probably more TODO things to come

Some benchmark of our internal service (6 endpoints):

- master branch - 158.54 seconds
- this branch - 53.4 sec in 6 worker threads

### DIFF
--- a/src/schemathesis/_hypothesis.py
+++ b/src/schemathesis/_hypothesis.py
@@ -3,7 +3,7 @@ import asyncio
 import re
 from base64 import b64encode
 from functools import partial
-from typing import Any, Callable, Dict, Generator, Optional
+from typing import Any, Callable, Dict, Generator, Optional, Union
 from urllib.parse import quote_plus
 
 import hypothesis
@@ -33,6 +33,15 @@ def create_test(
     if settings is not None:
         wrapped_test = settings(wrapped_test)
     return add_examples(wrapped_test, endpoint)
+
+
+def make_test_or_exception(
+    endpoint: Endpoint, func: Callable, settings: Optional[hypothesis.settings] = None, seed: Optional[int] = None
+) -> Union[Callable, InvalidSchema]:
+    try:
+        return create_test(endpoint, func, settings, seed=seed)
+    except InvalidSchema as exc:
+        return exc
 
 
 def get_original_test(test: Callable) -> Callable:
@@ -126,9 +135,7 @@ def get_case_strategy(endpoint: Endpoint) -> st.SearchStrategy:
             if value is not None:
                 if parameter == "path_parameters":
                     strategies[parameter] = (
-                        from_schema(value)
-                        .filter(filter_path_parameters)  # type: ignore
-                        .map(quote_all)  # type: ignore
+                        from_schema(value).filter(filter_path_parameters).map(quote_all)  # type: ignore  # type: ignore
                     )
                 elif parameter == "headers":
                     strategies[parameter] = from_schema(value).filter(is_valid_header)  # type: ignore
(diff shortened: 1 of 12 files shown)


src/schemathesis/runner/__init__.py
```diff
@@ -1,116 +1,155 @@
...
-) -> Generator[requests.Session, None, None]:
-    with requests.Session() as session:
-        if auth is not None:
-            session.auth = auth
-        session.headers["User-agent"] = USER_AGENT
-        if headers is not None:
-            session.headers.update(**headers)
-        yield session
 
+class Worker(threading.Thread):
```
Stranger6667: Maybe, default `Thread` will be easier, just with passing everything needed to the execution:

```
Thread(target = run, args = (...))
```

src/schemathesis/_hypothesis.py
```diff
@@ -126,9 +135,7 @@ def get_case_strategy(endpoint: Endpoint) -> st.SearchStrategy:
             if value is not None:
                 if parameter == "path_parameters":
                     strategies[parameter] = (
-                        from_schema(value)
-                        .filter(filter_path_parameters)  # type: ignore
-                        .map(quote_all)  # type: ignore
+                        from_schema(value).filter(filter_path_parameters).map(quote_all)  # type: ignore  # type: ignore
```
Stranger6667: Some weird artifact

src/schemathesis/runner/__init__.py
```diff
@@ -1,116 +1,155 @@
...
+        self.tasks_queue = tasks_queue
+        self.events_queue = events_queue
+        self.schema = schema
+        self.checks = checks
+        self.results = results
+        self.settings = settings
+
+    def run(self) -> None:
+        # TODO. catch hypothesis output - we should move it to the main thread
+        with get_session() as session:
```
Stranger6667: Missing auth & headers.

src/schemathesis/runner/__init__.py
```diff
@@ -1,116 +1,155 @@
...
+            # each call to `is_alive` of an alive worker waits for a lock
+            # iterations without waiting are too frequent and a lot of time will be spent on waiting for this locks
+            time.sleep(0.001)
+            is_finished = all(not worker.is_alive() for worker in workers)
+            while not events_queue.empty():
+                yield events_queue.get()
+
+    def _get_tasks_queue(self) -> Queue:
+        """All endpoints are distributed among all workers via a queue."""
+        endpoints = list(self.schema.get_all_endpoints())
```
Stranger6667: could be used in extend without calling `list` here

PR #323, opened 2019-12-11
Fix compatibility with Hypothesis >= 4.53.2

### LINKED ISSUE #322: New release of hypothesis breaks schemathesis
```
packages/schemathesis/models.py", line 9, in <module>
    from hypothesis.searchstrategy import SearchStrategy
ModuleNotFoundError: No module named 'hypothesis.searchstrategy'
```

`hypothesis` just released an hour ago...breaks this.

Release notes include:

>If you are depending on the internal API (which you shouldn't be,    and which we don't guarantee compatibility on) you may have to rename some   imports.

Stranger6667:
Closes #322 

It is compatible with the previous Hypothesis versions

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,11 @@ Changelog
 `Unreleased`_
 -------------
 
+Fixed
+~~~~~
+
+- Compatibility with Hypothesis >= 4.53.2. `#322`_
+
 `0.19.0`_ - 2019-12-02
 ----------------------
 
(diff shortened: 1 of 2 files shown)


kdelee:
Should you also add a minimum version to the dependency?

Stranger6667:
There is no need for the minimum version pin since it is compatible with the previous versions (tested on 4.39.3 and 4.34.0) 

I will add some more test builds later to verify this compatibility

Stranger6667:
Most probably I was relying on automatic PyCharm imports when I was developing that module - I didn't check that `searchstrategy` is internal

Stranger6667:
I will re-check though

kdelee:
Cool, thanks

Stranger6667:
Yep, re-checked Hypothesis code & rerun few more versions in tox

PR #328, opened 2019-12-12
Add `Case.verify_response` for running built-in checks against app's response

### LINKED ISSUE #319: How to check the response codes?
Hi! Thanks for such a useful tool!
I would like to understand if response codes are checked with an openapi scheme?

Stranger6667:
Closes #319

### DIFF
--- a/src/schemathesis/checks.py
+++ b/src/schemathesis/checks.py
@@ -1,22 +1,24 @@
-from typing import Callable, Tuple, Union
+from typing import TYPE_CHECKING, Callable, Tuple, Union
 
 import jsonschema
 import requests
 
-from ..models import TestResult
-from ..utils import WSGIResponse, are_content_types_equal
+from .utils import WSGIResponse, are_content_types_equal
+
+if TYPE_CHECKING:
+    from .models import Case
 
 GenericResponse = Union[requests.Response, WSGIResponse]  # pragma: no mutate
 
 
-def not_a_server_error(response: GenericResponse, result: TestResult) -> None:
+def not_a_server_error(response: GenericResponse, case: "Case") -> None:
     """A check to verify that the response is not a server-side error."""
     if response.status_code >= 500:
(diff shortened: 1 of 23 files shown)


src/schemathesis/models.py
```diff
@@ -11,6 +11,8 @@
 import werkzeug
 from hypothesis.strategies import SearchStrategy
 
+from schemathesis.checks import ALL_CHECKS
```
Stranger6667: Relative import

src/schemathesis/models.py
```diff
@@ -23,17 +25,30 @@
...
+    endpoint: "Endpoint" = attr.ib()  # pragma: no mutate
     path_parameters: Optional[PathParameters] = attr.ib(default=None)  # pragma: no mutate
     headers: Optional[Headers] = attr.ib(default=None)  # pragma: no mutate
     cookies: Optional[Cookies] = attr.ib(default=None)  # pragma: no mutate
     query: Optional[Query] = attr.ib(default=None)  # pragma: no mutate
     body: Optional[Body] = attr.ib(default=None)  # pragma: no mutate
     form_data: Optional[FormData] = attr.ib(default=None)  # pragma: no mutate
 
+    @property
+    def path(self) -> str:
```
Stranger6667: Check how the output will look like since some CLI output formatter is dependent on the Case attributes

src/schemathesis/checks.py
```diff
@@ -48,11 +50,11 @@ def content_type_conformance(response: GenericResponse, result: TestResult) -> N
     )
 
 
-def response_schema_conformance(response: GenericResponse, result: TestResult) -> None:
+def response_schema_conformance(response: GenericResponse, case: "Case") -> None:
```
Stranger6667: It is a backward-incompatible change - should be explicitly noted in the changelog

PR #337, opened 2019-12-18
Change hypothesis_jsonschema private module name

hlobit:
The commit https://github.com/Zac-HD/hypothesis-jsonschema/commit/1d98b386503b2291eab45d0290bf61c0a8761452 introduces a change in `hypothesis_jsonschema` internal structure that affects `schemathesis`.

It is just a patch, until something is found to avoid usage of the lib internals ?


Stranger6667:
Hello @hlobit :)

Cool, thank you for noticing this - there was no `hypothesis-jsonschema` release with these changes, and I'd like to use a public API if possible. Some compatibility shim should be used anyway to support `hypothesis-jsonschema` versions after & before these changes. Or we can pin `hypothesis-jsonschema` version with these changes when it will be out - not sure actually what would be the best way.

Maybe we can involve @Zac-HD into the discussion? What do you think about having such public API for registering string formats in `hypothesis-jsonschema`  itself?

hlobit:
> There was no `hypothesis-jsonschema` release with these changes

Yes there is 0.9.13 released today and that's how I noticed this :)

Zac-HD:
It's out as of about an hour ago, since I discovered an array canonicalisation bug. So requiring >=0.9.13 would work.

What's your use-case for this? I'm keen to keep my public api as small as possible, but cam grow it if needed!

Stranger6667:
For example, we have some custom format, e.g. for payment card numbers - and we declare it in the schema + use validation for it via `jsonschema` + `connexion`:

```
@draft4_format_checker.checks("card_number")
def is_card_number(value):
    # Validation based on Luhn algorithm 
```

So it is triggered for all input values with `format: "card_number"`. 
And it is very nice to generate valid card numbers to get deeper in the code with generated input

It is basically our use-case :)

Zac-HD:
I've opened Zac-HD/hypothesis-jsonschema#29 to track that and discuss the design :smile:

PR #340, opened 2020-01-08
Add tests for different Hypothesis versions

### LINKED ISSUE #324: Add builds for different Hypothesis versions
Follow up on #322 

We need to ensure compatibility. I propose to add 4.50.0, 4.40.0 at least to the build matrix

svtkachenko:
Closes #324

### DIFF
--- a/test/cli/test_commands.py
+++ b/test/cli/test_commands.py
@@ -5,6 +5,7 @@ from test.utils import HERE, SIMPLE_PATH
 import pytest
 import yaml
 from _pytest.main import ExitCode
+from importlib_metadata import version
 from hypothesis import HealthCheck, Phase, Verbosity
 from requests import Response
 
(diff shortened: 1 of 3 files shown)


Stranger6667:
Could you , please, add a condition for the list of available phases in the test for help output? So on hypothesis 4.40 the test won’t fail

test/cli/test_commands.py
```diff
@@ -70,12 +71,15 @@ def test_commands_version(cli):
...
         (
             ("run", "http://127.0.0.1", "--hypothesis-phases=explicit,first,second"),
-            'Error: Invalid value for "--hypothesis-phases": invalid choice(s): first, second. '
-            "Choose from explicit, reuse, generate, target, shrink",
+            'Error: Invalid value for "--hypothesis-phases": invalid choice(s): first, second. ' "Choose from {phases}",
         ),
     ),
 )
 def test_commands_run_errors(cli, args, error):
+    phases = "explicit, reuse, generate, target, shrink"
```
Stranger6667: I'd rather move it to the module level and use f-string

PR #344, opened 2020-01-10
Raise proper exception when the given schema is invalid

### LINKED ISSUE #308: Show understandable errors when the given schema is invalid
For example when a reference is not resolved, then it should be clear what happened, now it could be an `AttributeError`

svtkachenko:
Closes #308

### DIFF
--- a/test/test_schemas.py
+++ b/test/test_schemas.py
@@ -1,7 +1,11 @@
+from copy import deepcopy
+from test.utils import as_param
+
 import pytest
 from jsonschema import RefResolver
 
 import schemathesis
+from schemathesis.exceptions import InvalidSchema
 
 
 @pytest.mark.parametrize("base_path", ("/v1", "/v1/"))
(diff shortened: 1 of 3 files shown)


test/test_schemas.py
```diff
@@ -88,3 +91,15 @@ def test_resolving_multiple_files():
         },
         "xml": {"name": "User"},
     }
+
+
+@pytest.mark.parametrize("error_type", ("KeyError", "AttributeError", "RefResolutionError"))
+def test_schema_parsing_error(swagger_20, error_type):
+    if error_type == "KeyError":
+        swagger_20.raw_schema.pop("paths")
```
Stranger6667: This fixture is shared, you need to modify a copy

test/test_schemas.py
```diff
@@ -88,3 +91,15 @@ def test_resolving_multiple_files():
...
+    elif error_type == "AttributeError":
+        swagger_20.raw_schema["paths"] = {None: ""}
+    elif error_type == "RefResolutionError":
+        swagger_20.raw_schema = as_param({"$ref": "#/definitions/SimpleIntRef"})
+    with pytest.raises(InvalidSchema):
```
Stranger6667: Please, test the error message as well

PR #380, opened 2020-01-28
CLI option to fail after the first error

### LINKED ISSUE #378: CLI option to fail after the first error
It may be similar to pytest's one

svtkachenko:
Closes #378

### DIFF
--- a/src/schemathesis/cli/__init__.py
+++ b/src/schemathesis/cli/__init__.py
@@ -50,6 +50,9 @@ def schemathesis(pre_run: Optional[str] = None) -> None:
     "--checks", "-c", multiple=True, help="List of checks to run.", type=CHECKS_TYPE, default=DEFAULT_CHECKS_NAMES
 )
 @click.option(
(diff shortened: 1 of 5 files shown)


src/schemathesis/cli/__init__.py
```diff
@@ -129,6 +130,7 @@ def run(  # pylint: disable=too-many-arguments
     auth_type: str,
     headers: Dict[str, str],
     checks: Iterable[str] = DEFAULT_CHECKS_NAMES,
+    exitfirst: bool = False,
```
Stranger6667: you could provide the third argument to `click.option` as `"exit_first"` and then it will be named  here as `exit_first` - in this case we won't have different names for this entity

src/schemathesis/runner/__init__.py
```diff
@@ -22,6 +22,7 @@
 from ..schemas import BaseSchema
 from ..utils import WSGIResponse, capture_hypothesis_output, get_base_url
 from . import events
+from .events import Finished
```
Stranger6667: It seems to be not used

PR #419, opened 2020-02-12
chore(ci): Add Windows build

svtkachenko:


Stranger6667:
Could you, please, add some postfix to build names? Now Windows / Linux builds are indistinguishable from their names

PR #420, opened 2020-02-12
chore(ci): Add Windows build

### LINKED ISSUE #401: Test builds for Windows
Ref #399

svtkachenko:
Resolves #401

### DIFF
--- a/test/test_app.py
+++ b/test/test_app.py
@@ -1,3 +1,4 @@
+import platform
 import sys
 from pathlib import Path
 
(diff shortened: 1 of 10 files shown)


test/test_app.py
```diff
@@ -1,5 +1,6 @@
 import sys
 from pathlib import Path
+from platform import platform
```
Stranger6667: It should be only `import platform`

test/test_app.py
```diff
@@ -18,4 +23,7 @@ def test_app(testdir, aiohttp_unused_port, framework, expected):
     with testdir.tmpdir.join("stderr").open() as fd:
         stderr = fd.read()
     # Then it should start OK and emit debug logs
-    assert expected in stderr
+    if len(expected) == 2 and platform.system() == "Windows":
```
Stranger6667: You could do it on a module level and have consistent "expect" values:

```
if platform.system() == "Windows":
    AIOHTTP_OUTPUT = "DEBUG:asyncio:Using selector:"
else:
    AIOHTTP_OUTPUT = "DEBUG:asyncio:Using proactor: IocProactor"


@pytest.mark.parametrize(
    "framework, expected", (("flask", "INFO:werkzeug: * Running on"), ("aiohttp", AIOHTTP_OUTPUT))	    
)
```

test/test_dereferencing.py
```diff
@@ -1,3 +1,5 @@
+from platform import platform
```
Stranger6667: It should be only `import platform`

PR #430, opened 2020-02-26
fix: Error during `pytest` collection on objects that have custom `__getattr__` method and therefore pass `is_schemathesis` check

### LINKED ISSUE #429: schemathesis + mock.call break pytest collection
### The issue in a few words

Importing the `call` method from the `mock` package breaks pytest collection when schemathesis is installed, with the following error :

```
/usr/local/lib/python3.7/site-packages/_pytest/runner.py:237: in from_call
    result = func()
/usr/local/lib/python3.7/site-packages/_pytest/runner.py:257: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
/usr/local/lib/python3.7/site-packages/_pytest/python.py:431: in collect
    return super().collect()
/usr/local/lib/python3.7/site-packages/_pytest/python.py:370: in collect
    values.sort(key=lambda item: item.reportinfo()[:2])
/usr/local/lib/python3.7/site-packages/_pytest/python.py:370: in <lambda>
    values.sort(key=lambda item: item.reportinfo()[:2])
/usr/local/lib/python3.7/site-packages/_pytest/python.py:285: in reportinfo
    obj = self.obj
/usr/local/lib/python3.7/site-packages/_pytest/python.py:253: in obj
    self.own_markers.extend(get_unpacked_marks(self.obj))
/usr/local/lib/python3.7/site-packages/_pytest/mark/structures.py:249: in get_unpacked_marks
    return normalize_mark_list(mark_list)
/usr/local/lib/python3.7/site-packages/_pytest/mark/structures.py:264: in normalize_mark_list
    raise TypeError("got {!r} instead of Mark".format(mark))
E   TypeError: got pytestmark.mark instead of Mark
```

### Steps to reproduce

Clean environment (in a `python:3.7-slim-buster` docker image), python version 3.7.5.

Install the following requirements (includes `pytest`, `mock` and all the prerequisistes for schemathesis) :

[requirements.txt](https://github.com/kiwicom/schemathesis/files/4254756/requirements.txt)

Run the following test file with pytest :

```python
from mock import call

def test_schemathesis():
    assert True
```

The test should run without any issue. Then install schemathesis, and run it again. This should give the error above.

### Versions of components

The bug was reproduced using those versions of components :

OS: debian (slim python docker images)
python: 3.7.5, 3.6.10
schemathesis: v0.24.4 to v0.20.5 (there's another issue with prior versions).
mock: 4.0.0 and 2.0.0 (2.0 was released in 2016 so it's not a recent issue)
pytest: 5.3.5, 5.1.3, 4.6.9
Also tried using a previous pytest-mock version, still breaks.

Stranger6667:
Fixes #429

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,11 @@ Changelog
 `Unreleased`_
 -------------
 
+Fixed
+~~~~~
+
+- Error during ``pytest`` collection on objects that have custom ``__getattr__`` method and therefore pass ``is_schemathesis`` check. `#429`_
+
 `0.24.4`_ - 2020-02-22
 ----------------------
 
@@ -721,6 +726,7 @@ Fixed
 .. _0.3.0: https://github.com/kiwicom/schemathesis/compare/v0.2.0...v0.3.0
 .. _0.2.0: https://github.com/kiwicom/schemathesis/compare/v0.1.0...v0.2.0
 
+.. _#429: https://github.com/kiwicom/schemathesis/issues/429
 .. _#418: https://github.com/kiwicom/schemathesis/issues/418
 .. _#416: https://github.com/kiwicom/schemathesis/issues/416
 .. _#412: https://github.com/kiwicom/schemathesis/issues/412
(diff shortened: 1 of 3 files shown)


Stranger6667:
"Black" is failing for some reason, not important

PR #439, opened 2020-03-18
Fixes bug where an empty string would be accepted as a required path template variable

Hultner:
As discussed on gitter there's currently a bug where schemathesis will allow `""` as a valid input for a required path variable. This behaviour is incorrect and fixed with this patch.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,7 @@ Changelog
 Fixed
 ~~~~~
 
+- Filter invalid empty `""` path template variables. `#439`_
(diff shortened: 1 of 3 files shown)


Hultner:
> Could you, please, add a mention to our changelog? https://github.com/kiwicom/schemathesis/blob/master/docs/changelog.rst
> 
> Thanks

I'll do that.

Stranger6667:
The job failed initially because of some flaky test for concurrent execution, and then rerun failed because of some runner issue. I think it is ok to merge since all other builds were ok

PR #440, opened 2020-03-18
fix: Filter problematic strings with slashes

Hultner:
Filters problematic strings containing slashes or htmlencoded `%2F`.

These strings causes problems both in WSGI and ASGI based applications.

Follow up on https://github.com/kiwicom/schemathesis/pull/439 

Related issue: https://github.com/tiangolo/fastapi/issues/1132

### DIFF
--- a/src/schemathesis/_hypothesis.py
+++ b/src/schemathesis/_hypothesis.py
@@ -149,16 +149,28 @@ def get_case_strategy(endpoint: Endpoint) -> st.SearchStrategy:
 def filter_path_parameters(parameters: Dict[str, Any]) -> bool:
     """Single "." chars and empty strings "" are excluded from path by urllib3.
 
+    A path containing to "/" or "%2F" will lead to ambigious path resolution in
+    many frameworks and libraries, such behaviour have been observed in both
+    WSGI and ASGI applications.
+
     In this case one variable in the path template will be empty, which will lead to 404 in most of the cases.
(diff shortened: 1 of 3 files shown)


src/schemathesis/_hypothesis.py
```diff
@@ -149,16 +149,28 @@ def get_case_strategy(endpoint: Endpoint) -> st.SearchStrategy:
...
+        "/",
         "",
     )
 
-    return not any(value in path_parameter_blacklist for value in parameters.values())
+    return not any(
+        (
+            value in path_parameter_blacklist
+            or isinstance(value, str)
+            and path_parameter_blacklist[1] in value
```
Stranger6667: I propose to put slash symbol in a separate variable and reuse it, so the condition is not dependant on the index in `path_parameter_blacklist`
```
SLASH = "/"

...
def filter_path_parameters(...):
    path_parameter_blacklist = (".", SLASH, "")

    return not any(
        (
            value in path_parameter_blacklist
            or isinstance(value, str)
            and SLASH in value
        )
        for value in parameters.values()
    )
```

Hultner: Sounds very reasonable, will change

Hultner: Fixed this now.

Stranger6667:
Looks great to me! Thank you, Alexander! :)

PR #462, opened 2020-04-03
Store requests & responses

### LINKED ISSUE #379: Optionally store requests and responses that happen during the test
It might help a lot with debugging
We can store it in some "cassette"-style file (like VCR-py does).

For example:

```
$ schemathesis run --out cassette.yaml ...
```

Then `cassette.yaml` will contain all requests sent to the app and all responses received.

It might be useful for debugging, since putting them in the CLI output can bloat the output

Stranger6667:
Closes #379 

- [x] Add `command` to cassette
- [x] Documentation
- [x] Reuse prepared request to avoid double computation
- [x] CLI output about saved cassette file


jamescooke:
@Stranger6667 Looks good to me :+1: Happy to test anything / give feedback on output formats if that's helpful.

Stranger6667:
Hi @jamescooke ! Thanks for reviewing this :)

I was thinking that in CLI it might be used like this:

```
schemathesis run --store-network-log=cassette.json http://127.0.0.1/schema.yaml
```

And then there will be a file `cassette.json` with prettified JSON data like this:

```
{
    "meta": {...}  // some meta information, e.g. when the run happened, run parameters, etc.
    "interactions": [
        {
            "request": {...}   // serialized Case instance so it can be reproduced later
            "response": {...}  // serialized response
        }
    ]
}
```

I am not completely sure about what data could be helpful in such a cassette. What do you think?
And I wanted to start with a file, because displaying all requests / responses on stdout takes a lot of space and probably not convenient as is, I am not sure how to display it in a convenient form in CLI :(

Stranger6667:
Another idea is to create a `replay` command to replay such cassettes and maybe display diffs between old and new responses

jamescooke:
@Stranger6667 Thanks for sharing this idea - it looks good :+1: 

CLI invocation with `--store-network-log=` pointing at a JSON file seems like a good idea. I agree that outputting this in shell will be hard and not that helpful.

> I am not completely sure about what data could be helpful in such a cassette. What do you think?

My two main goals for this debugging info are:

* To see what request failed.
* To easily reproduce the failing call outside of Schemathesis - maybe with requests.

I've taken a stab at building out the cassette file with a passing example:

```json
{
    "meta": {
        "time": 1,
        "seed": 1,
        "cli": "schemathesis run --store-network-log=cassette.json http://127.0.0.1/schema.yaml"
    },
    "interactions": [
        {
            "result": "PASS",
            "request": {
                "method": "post",
                "args": [
                    "http://127.0.0.1/things"
                ],
                "kwargs": {
                    "data": {
                        "name": "Thing"
                    },
                    "headers": {
                        "user-agent": "my-app/0.0.1"
                    }
                }
            },
            "response": {
                "status_code": 201,
                "json": {
                    "result": "Thing created"
                },
                "headers": {
                    "content-encoding": "gzip"
                }
            }
        }
    ]
}
```

Just some notes about fields:

* `meta.cli` - I think it would be helpful to include this if possible to help someone reproduce the schemathesis run.

* `interactions[0].result` - if this is included, then it becomes really easy to find failing cases. For example with `jq`:  `jq '.["interactions"] | map(select(.result != "PASS"))' `

* `interactions[0].request` this should give enough information to reproduce a request. It looks a lot like a "Case" to me https://github.com/kiwicom/schemathesis/blob/cff8a8778ba34277ba8668fd02bcdb2aeb009470/src/schemathesis/models.py#L24-L33 Should it maybe be called "case"?

* `interactions[0].response.headers` would be the content of headers as seen by requests https://requests.readthedocs.io/en/master/user/quickstart/#response-headers

This would mean that failing requests could be easily reassembled with:

```py
import json
import requests
with open('cassette.json') as f:
    data = json.load(f)
for i in data['interactions']:
    if i['result'] != 'PASS':
        r = i['request']
        method = getattr(requests, r['method'])
        method(*r['args'], **r['kwargs']) 
```

Does that all make sense? I hope I've not pushed the scope of this idea further than you were expecting :crossed_fingers:

Stranger6667:
@jamescooke 
James, thank you so much for your detailed comment, I appreciate it :) It gives very helpful insights and does perfect sense for me :)

1. Great idea about having a command inside a cassette. I'd probably name it `command` to be more generic
2. :+1: I think we can even add this example with `jq` to our docs;
3. Yep, the renaming will make it more clear;
4. :+1: There is one problem with saving binary payload this way. I think that having base64-encoded will make it more usable in such cases (as `VCRpy` does). But it wouldn't make the code sample much more complex though. Anyway it will be handled in the `replay` command later

One extra thing. Probably having `id` field (with e.g. numeric id) can make debugging easier. For example, these IDs can be used in `replay` command to replay only needed requests, or this command could filter requests by status / another field. Probably it deserves another issue :)

I will work on implementation, thank you :)

Stranger6667:
Some updates from my side:

1. Using the JSON format is problematic because of binary data. With YAML it is easier since there is a `!!binary` tag and it is converted back and forth automatically. With JSON it will require custom encoder and extra data to understand which field inside `form_data` / `body` is in a binary format. At least the simplest option will be to start with YAML. Also, it is much easier to implement incremental writing with YAML to avoid delay at the end of the test session. It becomes noticeable on built-in test server with >1000 examples
2. Reproduction in `requests`. It might be easier to reuse the logic from the `Case` model since there are some extra conversions inside:

```
case = Case(**i['case'])
case.call()
```

Hultner:
> Some updates from my side:
> 
> 1. Using the JSON format is problematic because of binary data. With YAML it is easier since there is a `!!binary` tag and it is converted back and forth automatically. With JSON it will require custom encoder and extra data to understand which field inside `form_data` / `body` is in a binary format. At least the simplest option will be to start with YAML. Also, it is much easier to implement incremental writing with YAML to avoid delay at the end of the test session. It becomes noticeable on built-in test server with >1000 examples
> 2. Reproduction in `requests`. It might be easier to reuse the logic from the `Case` model since there are some extra conversions inside:
> 
> ```
> case = Case(**i['case'])
> case.call()
> ```

Would it be possible to use the StrictYAML subset? https://github.com/crdoconnor/strictyaml
To avoid some of the common [gotchas](https://hitchdev.com/strictyaml/features-removed/) with YAML.

Stranger6667:
@Hultner thank you for raising this point!

I am keen to have a stricter subset of YAML to avoid possible problems. As far as I see we might have a problem if we dump data and after loading, we might have something that is not equal to the original data. E.g. that "Norway problem" mentioned in the StrictYAML docs.

Unfortunately, it disallows explicit tags (`!!binary` in particular) which leads to the same complications as with JSON since arbitrary data must be encoded into a string first (it is `base64` with `pyyaml`). This feature is basically the reason to have YAML in the first place :( A workaround could be to convert everything `body` / `form_data` to base64 recursively and in this case, we can decode it. Or have type information in data, e.g. in some field. Or another solution to use some binary format like `msgpack` and convert its output to `base64`. Not sure about complexity / performance of this approach and how easy it will be to have incremental write.

As for dumping, we can control `pyyaml` events to have a particular style and we can avoid some ambiguity problems, but regarding binary data, we need to choose some approach and YAML will become not necessary at all / strictyaml can be used as is. I am bending to something like this:

```
{
    ...
    "body": {
        "raw": {"foo": 1},  // when there is no `bytes` data inside
        "encoded": "..."  // b64-encoded string if there are bytes inside. However not sure how to encode complex objects
    }
}
```

What do you think?

Hultner:
@Stranger6667 What's the problem with a base64 representation of the binary that doesn't exist in the pure binary representation? Isn't it just the binary data encoded into a 64-character space?

Otherwise I think it would be reasonable to use the strictYAML subset+binary, but use either use the pyyaml parser or patch the strictYAML-one, whichever is easier. 

Whichever way we go I think we should be careful with using YAML so we do it in a sensible and safe way.

Stranger6667:
There is no problem in this implementation itself, but it works only on `bytes`, however, there could be lists or dictionaries of arbitrary nesting in `form_data` or `body` with other types as well.

I'll try to give a more verbose example of my thoughts about this problem.

For example, this schema (modified version of [this](https://swagger.io/docs/specification/describing-request-body/file-upload/))

```
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                orderId:
                  type: integer
                title:
                  type: string
                fileName:
                  type: string
                  format: binary
```

So we might have this generated data:

```
body = {
    "orderId": 1,
    "userId": "foo",
    "fileName": b"x\xa6"  # this example is taken from `os.urandom(2)`
}
```

To serialize this to JSON we'll need to encode the value of `fileName` with `base64` and then decode these bytes with `utf8` (as one of the possible ways). Then this value will be not distinguishable from string types (in an arbitrary case). So, we need to keep this information somehow, I see these options:
- keep type info in the serialized output;
- some convention on the Schemathesis code level;
- Use a binary format for the whole body, e.g. serialize it with `messagepack`;

The first option is basically `!!binary` tag in YAML, otherwise we need to keep this information somehow embedded in the serialized data itself (not in a separate node type as tags are), either with a separate `type` field or with something similar, and it concerns me because we'll need to modify the structure with our own fields, then remove them on loading, etc. So quite a lot of actions to have a proper serialize - deserialize cycle.

In a simpler case we could just encode the whole body if it will be e.g. taken from the lower level like `urllib3` where it is already converted to bytes but in this way, it will be harder to implement and the serialized data will be dependent on `requests` and e.g. can't be reused with WSGI + other parts should be also taken from `urllib3` level, to have consistency in the serialized request structure.

The last option might work without putting extra info in the serialized data and there we can just serialize the whole body:

```
In [13]: base64.b64encode(msgpack.packb(body)).decode()                                                                                                                                                                                                                                                                          
Out[13]: 'g6dvcmRlcklkAaZ1c2VySWSjZm9vqGZpbGVOYW1lxAJ4pg=='
```

But I am not sure about, it is another convention with certain performance penalty (there will be separate calls for each request/response pair).

Hope that this will clarify my concerns about binary data in cassettes :)

> Otherwise I think it would be reasonable to use the strictYAML subset+binary, but use either use the pyyaml parser or patch the strictYAML-one, whichever is easier.

I will try to setup `pyyaml` with a proper style, I will post an update here, once there will be some results from my side.

> Whichever way we go I think we should be careful with using YAML so we do it in a sensible and safe way.

I agree with you and these YAML gotchas were completely out of my sight, thank you again for bringing this :)

Hultner:
@Stranger6667 I see, in that case it sounds like the last option using pyyaml with good configuration sounds like the least complex option? If so that's probably good.

Could make sense to take a look at [betamax](https://github.com/betamaxpy/betamax) as well and see what they're doing.

Stranger6667:
> I see, in that case it sounds like the last option using pyyaml with good configuration sounds like the least complex option?

Yep. I have a draft of it locally, but I still not satisfied with the performance, even with `CDumper` :/
However, maybe it will make sense to add another line to the output with the path to the saved cassette + maybe some small animation so it is visible that Schemathesis doesn't hang:

```
==== SUMMARY ====

...
Network log is saved to ...<path>

==== 6 passed, 1 failed in 0.09s ====
```

> Could make sense to take a look at betamax as well and see what they're doing

I was checking `VCRpy` and `betamax` and they are working at a lower level. In their case there is no need to preserve the original structure, the input is either a string or bytes that are saved under `string` / `base64_string` keys respectively. It is some kind of combination of things from my previous comment. 

Compatibility with VCR sounds interesting!

In general, it would be nice to have, but they are not compatible with WSGI clients and my idea was about having cassettes compatible with `Case` model, i.e. store generated data, that is not prepared for some particular use-case (requests / WSGI) at the moment. I need to think more about it

Stranger6667:
Btw, do you have some example of usage where VCR compatibility will be beneficial?

Hultner:
It seems like all vcr-format doesn't aim to maintain compatibility but here's some benefits on the top of my mind/5 min research
- Somewhat standardised format
    - Easier for external contributors to understand, while a new unique format will always need some cognitive overhead for developers external to the project
- Cassettes can be reused with other tools such as betamax or for instance the [Mr. Video](https://github.com/quidproquo/mr_video) frontend.
- Network effects, improvements in handling the format by other libraries can be reused/borrowed or act as inspiration for our implementation
- An established format is already somewhat battle proven and tested
- Speed, [VCRpy claims a 10x speedup](https://vcrpy.readthedocs.io/en/latest/installation.html#speed) over pyyaml

Another option could be looking into for instance the [HAR-format](https://en.wikipedia.org/wiki/HAR_(file_format)) which is also standardised. In general I'd prefer a standardised format if it doesn't require a ton of extra work up front since it offloads that piece of non-core expertise to an external entity.

Stranger6667:
Thank you for providing these points! I missed to write down some goals that I had in mind before I started the implementation and discussion here. I'll write them down below and will try to connect them to the current implementation details. Then I will share some more thoughts on the VCR compatibility and YAML.

**Supporting different protocols in the same output**
Currently, we support two protocols for communicating with the application under test and a third one is WIP (#512):
- HTTP
- WSGI
- ASGI (WIP)

The simplest common ground is `Case` model instance, which is exactly what was generated by Hypothesis, and it could be serialized to `HTTP` or `WSGI` compatible structure (which are consumed by `requests.request` and `werkzeug.Client.open` respectively.

**Reproducing requests exactly as they are**
Worth mentioning. 

**Performance**
It might not be an issue with a small number of generated examples and small schemas, but I usually run Schemathesis with `--hypothesis-max-examples=1000` or with a bigger number. 

Consider this config of our built-in server:

```
./test_server.sh 8081 --endpoints=success,failure,path_variable,multipart,recursive,multipart,payload,upload_file
```

And this straightforward cassettes use (in `cli.__init__.py:184` in the master branch):

```
with vcr.use_cassette("vcr.yaml"):
    prepared_runner = runner.prepare(
    ...
```

CLI command:

```
schemathesis run --hypothesis-max-examples=1000 --hypothesis-seed=1 --hypothesis-deadline=None http://127.0.0.1:8081/swagger.yaml
```

Note, these are average running times from 3 runs

The run times on my local machine (:
- 11.15s without "vcr"
- 54.10s with "vcr"  (x4.85)
And this one is already with C extension (the one you mention in the bullet list above).

Few versions ago all data was in the `Finished` event, but in experiments, it turned out that it is extremely slow to put everything in the last event (JSON representation was few dozens of megabytes in my examples), so I decided that we can distribute requests/responses data across `AfterExecution` events.

This idea could go deeper, so, we can record requests immediately when they happen, but there are some issues with that:
- There is already a communication channel for working with runner - events
- With events, there is more flexibility. It is not necessarily a file, it could be anything, in particular, I am worried about the web part, where these requests/responses can be displayed there;
- In the future, it will be nice to have the runner isolated, so it can be executed remotely and scaled easier;
There could be at least two approach to implement this:
- Patching, like VCR does, which won't work in subprocess on Windows (due to its forking semantic)
- Another queue to communicate + a separate thread to avoid blocking by the Hypothesis test, which is more complex;

Then, even with these events distribution, we need to write data to a file incrementally, otherwise, there is no point (it will be a big chunk in the end of the test run).
Now there is an incremental YAML writer based on the `pyyaml` events and its `CDumper`.
Run time - `12.33s` (x1.10)

This 10% penalty might be not that bad, but still, it can be better. And there will still be some delay after the last test. Having a separate thread will make it more smooth, so the main thread is not blocked, but still, performance win will be small, because of GIL and having CPU-bound test generation. However, since writing to a file and waiting for a response from the app is IO-bound then there could be some win, hard to say.

What I am actually considering is to write that logging handler completely in Rust, so it can do consuming events from a queue and writing them down to a file completely on the Rust side. It, again, depends, I don't know how much the FFI overhead will take (considering my experiments with PyO3's FFI it shouldn't be super big) and it depends on the number of requests/responses to process + there will be more issues with distributing a binary extension.

Maybe that delay, in the end, won't be that bad, I don't know will try to experiment and if it will be a reasonable delay (e.g. <500 ms on 1k requests) then maybe it can be OK.
However, it might be a reasonable price to have all requests stored.

**Having logs readable by humans if possible**
I think this one can be sacrificed if it will help to implement the goals above.

**VCR compatibility**

I think that it is nice to have some lower cognitive overhead indeed, but we have some more data, e.g. `elapsed` time which might be helpful to determine which requests were slow during the test run and there is some more helpful metadata (e.g. `command` to reproduce requests). I do think that having this metadata is more important than compatibility, and it will be quite straightforward to convert it by stripping down extensions.

Reusing these cassettes in different apps sounds cool! I can imagine that such cassettes can be used for some client-side code testing as well, but it is not clear for me yet how exactly they will be helpful. For example, we found some errors with Schemathesis and want to write regression test cases, then probably having mocks won't provide as much confidence as a real server response (e.g. a live one), not sure. But for displaying cassettes in some frontend it is nice to have indeed! :)

**YAML**
The more I dive into it the more I feel that using YAML without a schema will violate the first goal because we cannot have a schema for arbitrary structures (like in an example I provided in my previous comment), which leads me to use VCR-like format instead, where the body is always in bytes.

So, in the end, I think that there are more benefits in having VCR-compatible format (but with our extra data) than serializing `Case`:
- The body is always `bytes`, then we can always b64encode it and safely decode. No need for `!!binary` tag or some extra fields.
- In this way, we can have schemas for our cassettes so `strictyaml` can be used 
- JSON can be used as well
- Ideas from VCRpy / Betamax could be used in our serialization code

The serialization code won't be hard, it will follow the `requests` implementation (I guess that a lot can be used directly to prepare headers, body, etc)

However, in the future to implement `replay` we'll need to adapt this cassettes format to WSGI, which probably won't be a big issue (code from `werkzeug` could be used as an example)

Re: HAR. Not sure about this one, the standard is abandoned as stated on the W3C website, maybe there is a replacement but I think that having VCR compatibility will be fine and if there will be a demand we can add extra serialization format later

And again, @Hultner thank you for your comments and feedback, it gave me a lot of ideas and changed my view on certain aspects of this feature

Hultner:
@Stranger6667 Glad to be of help. 

From my point of view a 10% penalty is more then acceptable for a first version. I was thinking about the thread approach already before you mentioned it and I can't help but think that a separate process/sub-interpreter may be another viable approach, in that case we won't have to worry about the GIL but memory usage will be slightly higher. I'm thinking Erlang style simple message passing between the main process and a dedicated writer process, this will also enforce cleaner separation of concerns and may make it easier to implement a custom writer as you discuss.

Another alternative which we have not yet discussed is SQLLite, support is already built into python. It may not be a plaintext human readable format but it's a very common format that a lot of developers are familiar with and it's already heavily utilised on platforms like Android and iOS to store user data. That would also make it easier for user to use custom providers (following the DB-API), so that they for instance can store it in their own Postgres database or similar.

Stranger6667:
I tried a threaded approach now on the same endpoints and it gave me `11.42s` which is around `2.4%` of overhead, I believe it is ok :)

> I can't help but think that a separate process/sub-interpreter may be another viable approach

I think it could work, but with a bigger amount of data, so the overhead from a new process startup will be less than the performance gain

>  I'm thinking Erlang style simple message passing between the main process and a dedicated writer process, this will also enforce cleaner separation of concerns and may make it easier to implement a custom writer as you discuss.

Indeed! I'll submit my draft soon

Re: SQLLite. I think that it might be one of the output formats in the future (so the scope of this MR will remain smaller) and I like the idea, the implementation can be basically a different handler/writer process. 

> Have you measured where time is being spent?
My assumption was about serialization, but I didn't do profiling. Apparently, in HTTP-based tests the CPU-bound part was a lesser one since a separate writer thread improved performance of writing in ~4 times (assuming the time diffs for that extra log writing in different implementations). Having this amount of overhead I don't think that having a writer implemented in Rust will give much user-visible performance boost until some huge amount of data will be processed.

Here is a performance comparison for a built-in WSGI app:
- `7.69s` without saving logs
- `8.40s` with a thread writer

Command: 
```
PYTHONPATH=`pwd` schemathesis run --app=test.apps._flask.app:app --hypothesis-max-examples=1000 --hypothesis-seed=1 --hypothesis-deadline=None /swagger.yaml
```

It is around ~9.2% which is also ok from my pov.

jamescooke:
Thanks @Stranger6667 for your hard work on this. I agree with @Hultner that 2.4% performance hit is a tiny trade off for such a great feature.

Thanks @Hultner for your input on different formats. Binary requests were the elephant in the room when I was typing out an example JSON blob.

I'm looking forward to helping test this when you have a candidate :+1:

Stranger6667:
You are very welcome @jamescooke :)
I didn't expect that this feature will take that long :)
Now, the cassettes are compatible with VCR generally, and there some extensions from our side :) but the performance part changed.

For HTTP the same command now takes ~ `13.9s` which is around `22.45%`. I did some profiling and the biggest part goes from the serialization logic inside the runner. I assume that profiling is messing the numbers a bit since the profiling itself is not counted.

Here is a breakdown of `network_test` level that takes `98.4%` of time (stdout output takes the rest, and the sum of the breakdown should be ~`98.4%`). It is the main thread:
- 68.1% is Hypothesis's `wrapped_test`. Input generation inside and so on;
- 24.2% is `Case.call`, the network part;
- 6.0% is `store_requests_response`. and 5% of the total is spent in `requests.Session.prepare_request` inside this part.

From the writer thread:
- 87.2% - waiting for new events from the input queue;
- 12.1% - serialization of mapping (request or response)

The writer thread is probably fine with such breakdown, we definitely can squeeze some performance there with some Rust code, but probably it could be done later. The current logic is quite straightforward to port to [serde](https://github.com/serde-rs/serde). But still, with thread switching it is hard to predict the overall performance impact there, there will be some, but I am not expecting much.

As for the main thread. `Session.prepare_request` takes a lot, but for HTTP this part is already performed during the request sending. I'll try to reuse it and avoid these repeating computations

And with WSGI it went to ~`10s` which is around 30% overhead.

The above measurements are not really precise, but at least they give some picture of what is going on. What do you think?

What do you think about postponing JSON format for cassettes to a separate PR?

Stranger6667:
Ok, after reusing `response.request` the running time went down to `11.87s` which is ~6.5% overhead for HTTP interactions. Which is good from my pov. The WSGI case is still around 30%, but I don't see what we can do here for now. Will play with Rust implementation of writer thread this weekend. 
But I think it should be ok to have these numbers for the initial version.

Let me know what kind of metadata could be useful to add! (I am working on adding `command` to cassettes)

jamescooke:
> For HTTP the same command now takes ~ 13.9s which is around 22.45%.

I think even this is OK - certainly from my part. My expectation is that we (my current team) will only run with `--store-network-log` locally when trying to reproduce an error / bug.

> What do you think about postponing JSON format for cassettes to a separate PR?

Separating PRs is good for organisation. This already looks like a large internal change - so I would suggest hack and slash as required :+1: 

Personally I'm not concerned about the JSON format output - my main goal is to have a method to pick out a failing request after a failing Schemathesis run. I will have a look at checking out this branch and generating a cassette so I can have a look at the output. :eyes:

src/schemathesis/cli/cassettes.py
```diff
@@ -0,0 +1,190 @@
...
+            # In the beginning we write metadata and start `http_interactions` list
+            self.queue.put(Initialize())
+        if isinstance(event, events.AfterExecution):
+            # Seed is always present at this point, the original Optional[int] type is there because `TestResult`
+            # instance is created before `seed` is generated on the hypothesis side
+            seed = cast(int, event.result.seed)
+            self.queue.put(Process(status=event.status.name.upper(), seed=seed, interactions=event.result.interactions))
+        if isinstance(event, events.Finished):
+            self.queue.put(Finalize())
+            self.worker.join(1)
```
Stranger6667, line 52: Maybe we can have a named constant instead

src/schemathesis/cli/cassettes.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+@attr.s(slots=True)
+class Initialize:
+    """Start up, the first message to make preparations before proceeding the input data."""
+
+
+@attr.s(slots=True)
```
Stranger6667, line 60: Comments `# pragma: no mutate` are missing in many places

src/schemathesis/cli/cassettes.py
```diff
@@ -0,0 +1,190 @@
...
+    """Emit scalar values as strings.
+
+    It is required to avoid possible issues with default YAML parsing.
+    For example "Norway problem", where `- no` will be parsed to `[False]`, but we have strings everywhere
+    therefore we need `- 'no'` and ['no'].
+    """
+
+    def serialize_node(self, node: yaml.Node, parent: Optional[yaml.Node], index: int) -> None:
+        """Types are mostly ignored because it reproduces behavior of the original serializer."""
+        alias = self.anchors[node]
```
Stranger6667, line 84: A comment about the origin of this code will be helpful

src/schemathesis/cli/cassettes.py
```diff
@@ -0,0 +1,190 @@
...
+    def key(name: str) -> yaml.ScalarEvent:
+        """Default style for mapping keys is without quotes."""
+        return yaml.ScalarEvent(anchor=None, tag=None, implicit=(True, True), value=name)
+
+    def value(_value: str) -> yaml.ScalarEvent:
+        """Default style for mapping values is with quotes."""
+        return yaml.ScalarEvent(anchor=None, tag=None, implicit=(False, True), value=_value)
+
+    def serialize_mapping(name: str, data: Dict[str, Any]) -> None:
+        emit(yaml.ScalarEvent(anchor=None, tag=None, implicit=(True, True), value=name))
```
Stranger6667, line 150: `key` function can be reused

src/schemathesis/cli/cassettes.py
```diff
@@ -0,0 +1,190 @@
...
+                        key("id"),
+                        value(str(current_id)),
+                        key("status"),
+                        value(item.status),
+                        key("seed"),
+                        value(str(item.seed)),
+                        key("recorded_at"),
+                        value(interaction.recorded_at),
+                    )
+                    # These mappings should be also handled more strictly
```
Stranger6667, line 181: Obsolete comment

src/schemathesis/models.py
```diff
@@ -126,22 +130,24 @@ def call(
             session.close()
         return response
 
-    def as_werkzeug_kwargs(self) -> Dict[str, Any]:
+    def as_werkzeug_kwargs(self, headers: Optional[Dict[str, str]] = None) -> Dict[str, Any]:
         """Convert the case into a dictionary acceptable by werkzeug.Client."""
-        headers = self.headers
+        final_headers = self.headers.copy() if self.headers is not None else {}
```
Stranger6667, line 135: Probably an extra test for handling of multiple cookies will be nice to have

src/schemathesis/models.py
```diff
@@ -259,13 +262,135 @@ class Check:
...
+
+    @classmethod
+    def from_prepared_request(cls, prepared: requests.PreparedRequest) -> "Request":
+        """A prepared request version is already stored in `requests.Response`."""
+        body = getattr(prepared, "raw", getattr(prepared, "body", None))
+
+        if not body:
+            body = ""
+
+        if hasattr(body, "encode"):
```
Stranger6667, line 291: do we always have it? if yes, then we can encode it right away. And also, it will be nice to test the handling of bodies that are not utf-8 encodable

src/schemathesis/models.py
```diff
@@ -259,13 +262,135 @@ class Check:
...
+class Response:
+    """Unified response data."""
+
+    status_code: int = attr.ib()  # pragma: no mutate
+    message: str = attr.ib()  # pragma: no mutate
+    headers: Dict[str, str] = attr.ib()  # pragma: no mutate
+    body: str = attr.ib()  # pragma: no mutate
+    encoding: str = attr.ib()  # pragma: no mutate
+    http_version: str = attr.ib()  # pragma: no mutate
+    elapsed: float = attr.ib()  # pragma: no mutate
```
Stranger6667, line 324: Seems like not used in cassettes and not mentioned in the docs

test/cli/test_cassettes.py
```diff
@@ -0,0 +1,21 @@
...
+@pytest.mark.endpoints("success", "upload_file")
+def test_store_cassette(cli, schema_url, tmp_path):
+    file_ = tmp_path / "output.yaml"
+    result = cli.run(schema_url, f"--store-network-log={file_}", "--hypothesis-max-examples=2")
+    assert result.exit_code == ExitCode.OK
+    with file_.open() as fd:
+        cassette = yaml.safe_load(fd)
+    assert len(cassette["http_interactions"]) == 3
+    assert cassette["http_interactions"][0]["id"] == "0"
+    assert cassette["http_interactions"][0]["status"] == "SUCCESS"
```
Stranger6667, line 17: check more details from the cassette

test/runner/test_runner.py
```diff
@@ -107,10 +104,10 @@ def test_execute_base_url_found(base_url, schema_url, app):
     assert_incoming_requests_num(app, 3)
 
 
-def test_execute(args):
+def test_execute(request, args):
```
Stranger6667, line 107: could be reverted back

Stranger6667:
Thank you for insights, @jamescooke ! Hopefully, it will be shipped soon :)

Also, should we care about headers' sanitization, etc?

PR #476, opened 2020-04-04
Subprocess running mode

Stranger6667:
The goal is to be able to run `runner` in a subprocess while still having the same interface with an event generator. Few things should be done to achieve it:

- [x] Make all events serializable with `pickle`
- [x] Make all runner input values also serializable with `pickle` (`app` is not)
- [x] Connect a subprocess with the main one via `multiprocessing.Queue`
- [x] Parametrize all tests to make sure, that all features work in both modes

**Update**:
Serialization of all small bits of events is quite annoying :( Tracebacks, exceptions, apps, hooks need special care. Most of the logic works out of the box, though. Maybe as the first step, it will be better to make events independent from `schema`, `endpoint`, etc and store what is needed right there, without using some logic from `BaseSchema` and other classes, store serialized tracebacks, errors as strings, etc so nothing complex is sent between parent / child processes. It probably will help in the future, e.g. it will be easier to implement the usage of CLI to control remote nodes:
- [x] Remove all usage of `ExecutionEvent.schema` in CLI, use static data;
- [x] The same for `endpoint`

### DIFF
--- a/src/schemathesis/cli/__init__.py
+++ b/src/schemathesis/cli/__init__.py
@@ -1,16 +1,13 @@
 import traceback
-from contextlib import contextmanager
 from enum import Enum
-from typing import Any, Callable, Dict, Generator, Iterable, List, Optional, Tuple, Union, cast
+from typing import Callable, Dict, Generator, Iterable, List, Optional, Tuple, Union, cast
 
 import click
 import hypothesis
 import requests
-from requests import exceptions
 
 from .. import checks as checks_module
-from .. import models, runner, utils
-from ..exceptions import HTTPError
+from .. import models, runner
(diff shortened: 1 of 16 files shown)


Stranger6667:
I think, that I rather split these two commits into separate PRs. Events still need to be improved - even though they are serializable now, most of the data is concentrated in the `Finished` event.
Testing on the built-in test server these are sizes (in bytes) of different types of events:
- `Initialized` (173)
- `BeforeExecution` (39-53)
- `AfterExecution` (38)
- `Finished` (103699)
It contains all test results, and in a case of some big test runs, it might have megabytes of data. But we can spread this information across other events and CLI need to store it on its side

Issue #427, opened 2020-02-23
Export CLI test results to JUnit.xml


Stranger6667:
Propose a structure & implement it after discussion.
Maybe it will be possible to reuse `--junitxml` from `pytest`.

svtkachenko:
Closed via #528 and #534

PR #528, opened 2020-04-24
Add output to JunitXML

svtkachenko:
Implements #427


src/schemathesis/cli/__init__.py
```diff
@@ -116,6 +116,7 @@ def schemathesis(pre_run: Optional[str] = None) -> None:
     type=click.IntRange(1),
 )
 @click.option("--validate-schema", help="Enable or disable validation of input schema.", type=bool, default=True)
+@click.option("--junit-xml", help="Create junit-xml style report file at given path", type=str)
```
Stranger6667, line 119: Would it be more convenient with `click.File`?

src/schemathesis/cli/__init__.py
```diff
@@ -235,16 +237,20 @@ class OutputStyle(Enum):
 
     default = output.default.DefaultOutputStyleHandler
     short = output.short.ShortOutputStyleHandler
+    junit_xml = output.junitxml.JunitXMLHandler
```
Stranger6667, line 240: It shouldn't be here, these are only for output handlers. JUnit should go to a file

src/schemathesis/cli/__init__.py
```diff
@@ -235,16 +237,20 @@ class OutputStyle(Enum):
...
     prepared_runner: Generator[events.ExecutionEvent, None, None],
     workers_num: int,
     show_errors_tracebacks: bool,
     store_network_log: Optional[click.utils.LazyFile],
+    junit_xml: Optional[str],
 ) -> None:
     """Execute a prepared runner by drawing events from it and passing to a proper handler."""
     handlers = [get_output_handler(workers_num)]
+    if junit_xml is not None:
+        handlers = [OutputStyle.junit_xml.value()]
```
Stranger6667, line 253: It shouldn't replace the output handler, they should work both

src/schemathesis/cli/__init__.py
```diff
@@ -235,16 +237,20 @@ class OutputStyle(Enum):
...
     short = output.short.ShortOutputStyleHandler
+    junit_xml = output.junitxml.JunitXMLHandler
 
 
 def execute(
     prepared_runner: Generator[events.ExecutionEvent, None, None],
     workers_num: int,
     show_errors_tracebacks: bool,
     store_network_log: Optional[click.utils.LazyFile],
+    junit_xml: Optional[str],
```
Stranger6667, line 248: I'd also recommend adding this value to `ExecutionContext`, so te output handler can show that JUnit report was saved to some file. The handler itself should be in the handlers list before the output handler

src/schemathesis/cli/output/junitxml.py
```diff
@@ -0,0 +1,46 @@
+import platform
```
Stranger6667, line 1: Could you, please, move it out from `output` package to `cli`? these ones are only for stdout

src/schemathesis/cli/output/junitxml.py
```diff
@@ -0,0 +1,46 @@
...
+from junit_xml import TestSuite, TestCase, to_xml_report_string
+
+from ..handlers import EventHandler, ExecutionContext
+from . import default
+from ...models import Status
+from ...runner import events
+
+
+class JunitXMLHandler(EventHandler):
+    test_cases = []
```
Stranger6667, line 14: It should be `attr.ib(factory=list)`

src/schemathesis/cli/output/junitxml.py
```diff
@@ -0,0 +1,46 @@
...
+import click
+from junit_xml import TestSuite, TestCase, to_xml_report_string
+
+from ..handlers import EventHandler, ExecutionContext
+from . import default
+from ...models import Status
+from ...runner import events
+
+
+class JunitXMLHandler(EventHandler):
```
Stranger6667, line 13: Please, add `@attr.s(slots=True)` decorator

src/schemathesis/cli/output/junitxml.py
```diff
@@ -0,0 +1,46 @@
...
+    test_cases = []
+    start_time = None
```
Stranger6667, line 15: Please, use `attr.ib(default=None)`

src/schemathesis/cli/output/junitxml.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def handle_event(self, context: ExecutionContext, event: events.ExecutionEvent) -> None:
+        if isinstance(event, events.Initialized):
+            self.start_time = event.start_time
+        if isinstance(event, events.BeforeExecution):
+            self.test_cases.append(TestCase(f"{event.method} {event.path}",))
```
Stranger6667, line 21: That comma can be removed

src/schemathesis/cli/output/junitxml.py
```diff
@@ -0,0 +1,46 @@
...
+                    if check.message:
+                        message = f"{idx}. {check.message}"
+                    else:
+                        message = None
+                self.test_cases[-1].add_failure_info(message=message)
+            if event.status == Status.error:
+                self.test_cases[-1].add_error_info(message=event.result.errors[-1].exception)
+        if isinstance(event, events.Finished):
+            ts = [TestSuite("schemathesis", test_cases=self.test_cases, hostname=platform.node(),)]
+            click.echo("------------XML-----------")
```
Stranger6667, line 37: Stdout handling should still happen in `default` / `short` handlers, this handler should only save data to a file

src/schemathesis/cli/output/junitxml.py
```diff
@@ -0,0 +1,46 @@
...
+            click.echo(to_xml_report_string(ts, prettyprint=True))
+            status_code = 0
+            if event.has_errors or event.has_failures:
+                status_code = 1
+            raise click.exceptions.Exit(status_code)
```
Stranger6667, line 42: It should not be handled here. it is already handled in the stdout handlers (and will be moved out from there in the future)

src/schemathesis/runner/events.py
```diff
@@ -62,13 +62,20 @@ class AfterExecution(ExecutionEvent):
     # Endpoint test status - success / failure / error
     status: Status = attr.ib()  # pragma: no mutate
     result: SerializedTestResult = attr.ib()  # pragma: no mutate
+    # Test running time
+    elapsed_time: float = attr.ib()
```
Stranger6667, line 66: Please, add `# pragma: no mutate` comment

src/schemathesis/runner/impl/core.py
```diff
@@ -114,7 +115,10 @@ def run_test(
         test, "_hypothesis_internal_use_generated_seed", None
     )
     results.append(result)
-    yield events.AfterExecution.from_result(result=result, status=status, hypothesis_output=hypothesis_output)
+    test_elapsed_time = time.monotonic() - test_start_time
```
Stranger6667, line 118: I think that it will be better to measure the test running time without other lines in this function, so only wrapping the test execution line should be ok (before the test and after `except` parts).

src/schemathesis/cli/context.py
```diff
@@ -22,3 +22,4 @@ class ExecutionContext:
     terminal_size: os.terminal_size = attr.ib(factory=shutil.get_terminal_size)  # pragma: no mutate
     results: List[SerializedTestResult] = attr.ib(factory=list)  # pragma: no mutate
     cassette_file_name: Optional[str] = attr.ib(default=None)  # pragma: no mutate
+    junit_xml_file: Optional[str] = attr.ib(default=None)
```
Stranger6667, line 25: Missing `pragma` comment

src/schemathesis/cli/output/default.py
```diff
@@ -319,6 +319,8 @@ def handle_finished(context: ExecutionContext, event: events.Finished) -> None:
     display_failures(context, event)
     display_application_logs(context, event)
     display_statistic(context, event)
+    if context.junit_xml_file:
```
Stranger6667, line 322: I think it can be in `display_statistic`, in a similar way as cassettes feature does

docs/usage.rst
```diff
@@ -71,3 +71,14 @@ Check payload in requests to ``/api/upload_file``:
...
+Export CLI test results to JUnit.xml
+------------------------------------
+
+It is possible to export test results to format, acceptable by such tools as Jenkins.
+
+.. code:: bash
+
+    $ schemathesis run --junit-xml=/path/junit.xml
+
+This command will create an XML at given path.
```
Stranger6667, line 84: Could you provide some minimal output example?

src/schemathesis/cli/junitxml.py
```diff
@@ -0,0 +1,42 @@
...
+from click.utils import LazyFile
+from junit_xml import TestCase, TestSuite, to_xml_report_file
+
+from ..models import Status
+from ..runner import events
+from .handlers import EventHandler, ExecutionContext
+from .output import default
+
+
+@attr.s(slots=True)
```
Stranger6667, line 14: Missing `# pragma: no mutate` here and after each attribute

src/schemathesis/cli/junitxml.py
```diff
@@ -0,0 +1,42 @@
...
+class JunitXMLHandler(EventHandler):
+    file_handle: LazyFile = attr.ib()
+    test_cases: List = attr.ib(factory=list)
+    start_time: Optional[float] = attr.ib(default=None)
+
+    def handle_event(self, context: ExecutionContext, event: events.ExecutionEvent) -> None:
+        if isinstance(event, events.Initialized):
+            self.start_time = event.start_time
+        if isinstance(event, events.BeforeExecution):
+            self.test_cases.append(TestCase(f"{event.method} {event.path}"))
```
Stranger6667, line 24: I think it could be done in `AfterExecution` branch below since this data is available there. To work on the same item without using `test_cases[-1]`.

src/schemathesis/cli/junitxml.py
```diff
@@ -0,0 +1,42 @@
...
+                        message = f"{idx}. {check.message}"
+                    else:
+                        message = None
+                self.test_cases[-1].add_failure_info(message=message)
+            if event.status == Status.error:
+                self.test_cases[-1].add_error_info(
+                    message=event.result.errors[-1].exception, output=event.result.errors[-1].exception_with_traceback
+                )
+        if isinstance(event, events.Finished):
+            ts = [TestSuite("schemathesis", test_cases=self.test_cases, hostname=platform.node())]
```
Stranger6667, line 41: could you, please, name the variable `test_suite`?

src/schemathesis/cli/output/default.py
```diff
@@ -224,25 +224,29 @@ def display_statistic(context: ExecutionContext, event: events.Finished) -> None
...
+    if total:
+        display_checks_statistics(total)
+
+    if context.cassette_file_name:
+        click.echo()
+        category = click.style("Network log", bold=True)
+        click.secho(f"{category}: {context.cassette_file_name}")
+
+    if context.junit_xml_file:
+        display_section_name(f"generated xml file {context.junit_xml_file}", separator="-")
```
Stranger6667, line 237: "generated" -> "Generated" to be more consistent with the message above.
"xml" -> "XML"

Stranger6667: Also I think it will be better to have this line without separators around it, but just a single line like "Network log" above

src/schemathesis/cli/junitxml.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def handle_event(self, context: ExecutionContext, event: events.ExecutionEvent) -> None:
+        if isinstance(event, events.Initialized):
+            self.start_time = event.start_time
+        if isinstance(event, events.BeforeExecution):
+            self.test_cases.append(TestCase(f"{event.method} {event.path}"))
+        if isinstance(event, events.AfterExecution):
+            self.test_cases[-1].elapsed_sec = event.elapsed_time
+            if event.status == Status.failure:
+                checks = default._get_unique_failures(event.result.checks)
```
Stranger6667, line 28: It would be better to have `_get_unique_failures` outside of both handlers, so `junitxml` will not be dependent on `default`

src/schemathesis/cli/handlers.py
```diff
@@ -9,3 +13,15 @@ def handle_event(self, context: ExecutionContext, event: events.ExecutionEvent)
     def shutdown(self) -> None:
         # Do nothing by default
         pass
+
+
+def _get_unique_failures(checks: List[SerializedCheck]) -> List[SerializedCheck]:
```
Stranger6667, line 18: Since it is used in other modules, I think the leading underscore in the function name is not needed

src/schemathesis/cli/junitxml.py
```diff
@@ -7,25 +7,23 @@
...
+    test_cases: List = attr.ib(factory=list)  # pragma: no mutate
+    start_time: Optional[float] = attr.ib(default=None)  # pragma: no mutate
 
     def handle_event(self, context: ExecutionContext, event: events.ExecutionEvent) -> None:
         if isinstance(event, events.Initialized):
             self.start_time = event.start_time
-        if isinstance(event, events.BeforeExecution):
-            self.test_cases.append(TestCase(f"{event.method} {event.path}"))
         if isinstance(event, events.AfterExecution):
+            self.test_cases.append(TestCase(f"{event.result.method} {event.result.path}"))
```
Stranger6667, line 23: Could you, please, create a variable here and `append` at the end of the branch to avoid that list access at all?

src/schemathesis/cli/junitxml.py
```diff
@@ -38,5 +36,5 @@ def handle_event(self, context: ExecutionContext, event: events.ExecutionEvent)
                     message=event.result.errors[-1].exception, output=event.result.errors[-1].exception_with_traceback
                 )
         if isinstance(event, events.Finished):
-            ts = [TestSuite("schemathesis", test_cases=self.test_cases, hostname=platform.node())]
-            to_xml_report_file(file_descriptor=self.file_handle, test_suites=ts, prettyprint=True)
+            testsuites = [TestSuite("schemathesis", test_cases=self.test_cases, hostname=platform.node())]
```
Stranger6667, line 39: "testsuites" -> "test_suites" 
So the variable has the same name as the keyword argument in `to_xml_report_file`

src/schemathesis/cli/junitxml.py
```diff
@@ -0,0 +1,41 @@
...
+    test_cases: List = attr.ib(factory=list)  # pragma: no mutate
+    start_time: Optional[float] = attr.ib(default=None)  # pragma: no mutate
+
+    def handle_event(self, context: ExecutionContext, event: events.ExecutionEvent) -> None:
+        if isinstance(event, events.Initialized):
+            self.start_time = event.start_time
+        if isinstance(event, events.AfterExecution):
+            test_case = TestCase(f"{event.result.method} {event.result.path}")
+            test_case.elapsed_sec = event.elapsed_time
+            self.test_cases.append(test_case)
```
Stranger6667, line 25: Could you, please, append at the end of this branch to avoid list access at all?

src/schemathesis/cli/junitxml.py
```diff
@@ -0,0 +1,41 @@
...
+    file_handle: LazyFile = attr.ib()  # pragma: no mutate
+    test_cases: List = attr.ib(factory=list)  # pragma: no mutate
+    start_time: Optional[float] = attr.ib(default=None)  # pragma: no mutate
+
+    def handle_event(self, context: ExecutionContext, event: events.ExecutionEvent) -> None:
+        if isinstance(event, events.Initialized):
+            self.start_time = event.start_time
+        if isinstance(event, events.AfterExecution):
+            test_case = TestCase(f"{event.result.method} {event.result.path}")
+            test_case.elapsed_sec = event.elapsed_time
```
Stranger6667, line 24: It can be added in the constructor

PR #547, opened 2020-05-02
feat: A new generic hooks system

Stranger6667:
Why a new hook system?

- Isolate & unify hook registration logic
- A more generic hooks execution logic
- Validation of registered hooks
- Register multiple hooks per name - it will allow implementing fixups. Otherwise, hooks are re-assigned and fixups (if they will use hooks) might be rewritten by user's hooks
- A more generic naming structure, now we have only hooks named by request parts names

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -1,0 +1,81 @@
+.. customization:
+
+Customization
+=============
+
+Often you need to modify certain aspects of Schemathesis behavior, adjust data generation, modify requests before
+sending, and so on. Schemathesis offers a hook mechanism which is similar to the pytest's one.
+
+Basing on the scope of the changes there are three levels of hooks:
+
+- Global. These hooks applied to all schemas in the test run;
+- Schema-local. Applied only for specific schema instance;
+- Test function specific. Applied only for a specific test function;
+
+To register a new hook function you need to use special decorators - ``register`` for global and schema-local hooks and ``apply`` for test-specific ones:
+
+.. code:: python
+
+    import schemathesis
+
+    @schemathesis.hooks.register
+    def before_generate_query(context, strategy):
+        return strategy.filter(lambda x: x["id"].isdigit())
+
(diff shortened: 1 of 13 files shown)


docs/customization.rst
```diff
@@ -0,0 +1,72 @@
...
+- Test function specific. Applied only for a specific test function;
+
+To register a new hook function you need to use special decorators - ``register`` for global and schema-local hooks and ``apply`` for test-specific ones:
+
+.. code:: python
+
+    import schemathesis
+
+    @schemathesis.hooks.register
+    def before_generate_query(strategy, context):
```
Stranger6667, line 25: Probably context should be passed as the first argument since it will be more obvious - it will always present in all hooks and on the first position

src/schemathesis/schemas.py
```diff
@@ -64,7 +64,7 @@ class BaseSchema(Mapping):
     endpoint: Optional[Filter] = attr.ib(default=None)  # pragma: no mutate
     tag: Optional[Filter] = attr.ib(default=None)  # pragma: no mutate
     app: Any = attr.ib(default=None)  # pragma: no mutate
-    hooks: Dict[HookLocation, Hook] = attr.ib(factory=dict)  # pragma: no mutate
+    hooks: HookDispatcher = attr.ib(factory=HookDispatcher)
```
Stranger6667, line 67: Missing `# pragma: no mutate`

src/schemathesis/hooks.py
```diff
@@ -28,16 +20,170 @@ def warn_deprecated_hook(hook: Hook) -> None:
...
 
-def register(place: str, hook: Hook) -> None:
-    warn_deprecated_hook(hook)
-    key = HookLocation[place]
-    GLOBAL_HOOKS[key] = hook
+@attr.s(slots=True)  # pragma: no mutate
+class HookContext:
+    """A context that is passed to some hook functions."""
+
+    endpoint: Endpoint = attr.ib()
```
Stranger6667, line 27: Missing pragma

Issue #546, opened 2020-05-01
Get operations by operationId


dongfangtianyu:
[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#fixed-fields-8](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#fixed-fields-8)
>Tools and libraries MAY use the operationId to uniquely identify an operation

sometime，we need retest a specified API, using operationId will be more intuitive and efficient than using filtering.
for example:
```
  /todo:
    get:
      tags:
        - Todo
      summary: Todo List
      operationId: todo_list_todo_get
      responses:
        '200':
          description: Successful Response'
        '404':
          description: Not Found
    post:
      tags:
        - Todo
      summary: Create Todo
      operationId: todo_post_todo_post
```
Something like:
```python
@schema.parametrize(method="post", endpoint="/todo")                                                 
def test_no_server_errors(case):
    # do something with the case
    pass
```
VS
```python
@schema.parametrize(operationId="todo_post_todo_post")                                                
def test_no_server_errors(case):
    # do something with the case
    pass
```

Stranger6667:
Hi @dongfangtianyu 

It is a great idea :) I'll work on it

dongfangtianyu:
Hi @Stranger6667 

I am trying to done this issues, but I have some problems and need help.

First, I added a parameter to parameterize()
```python
def parametrize(
        self,
        method: Optional[Filter] = NOT_SET,
        endpoint: Optional[Filter] = NOT_SET,
        tag: Optional[Filter] = NOT_SET,
        operationId: Optional[Filter] = NOT_SET,      ############# there
        validate_schema: Union[bool, NotSet] = NOT_SET,
    )
```
When  run pre-commit, I see message: ` too many arguments (6 / 5) (too many arguments)`
May be something wrong. Do you have any ideas to remind me?

Stranger6667:
Hi @dongfangtianyu,

Sure! you need to add `# pylint: disable=too-many-arguments` on the first line. Probably we'll need to reconsider this pylint rule. It will look something like this:

```
def parametrize(  # pylint: disable=too-many-arguments
        self,
        method: Optional[Filter] = NOT_SET,
        endpoint: Optional[Filter] = NOT_SET,
        tag: Optional[Filter] = NOT_SET,
        operationId: Optional[Filter] = NOT_SET,      ############# there
[...]

Issue #500, opened 2020-04-14
Improve documentation


Stranger6667:
Based on #32 and discussion with @Hultner 

- [x] Have a smaller quickstart
- [x] Link to docs in the top of the page (not only badge)
- [x] Move documentation content to a separate document
- [ ] Share roadmap info
- [ ] Examples of errors that could be found
- [x] Details behavior of available checks
- [ ] Possible workflows for Schemathesis. As a part of a test suite, CI step, CLI, etc
- [ ] [Usage](https://github.com/untitaker/pytest-subtesthack) of `subtesthack` for pytest fixtures.
- [x] Add info about using regex in filters

We also can take inspiration from [Hypothesis docs](https://hypothesis.readthedocs.io/en/latest/quickstart.html)

louisguitton:
+1 to adding the https://schemathesis.readthedocs.io/en/stable/ link to the github repo itself

Stranger6667:
@louisguitton 

I moved the link slightly upper, hope that now it is more visible in the readme :) Let me know if I can provide more info on areas, that are interesting for you

louisguitton:
hey @Stranger6667 nice thank you. I meant using the github feature of "URL to your project" if you see what I mean
![image](https://user-images.githubusercontent.com/7823843/88787663-81b96f80-d194-11ea-80e9-2c160835235a.png)

Stranger6667:
Oh, I see. Unfortunately, I don't have admin access to this repo anymore. Maybe @aexvir could help with putting "https://schemathesis.readthedocs.io/en/stable/" to the "about" section?

aexvir:
Added the info in the about section, and also gave you some permissions @Stranger6667 🙂 
Let me know if it's enough or if you need a higher access level

Stranger6667:
Thank you @aexvir ! The permission level is OK :)

PR #554, opened 2020-05-04
add filter parameter `operationId` 

dongfangtianyu:
trying to done  #546 
- schema.parametrize0 add new parameter `operationId`， `operationId` can is str or list
- when the schema is instantiated, a dict (schema.schema_map) record the mapping relationship between operationid and endpoint
- when call get_all_endpoints,  if has operationId than get endpoint from schema_map

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,6 +10,7 @@ Added
 ~~~~~
 
 - Display a suggestion to disable schema validation on schema loading errors in CLI. `#531`_
+- ``operation_id`` argument can get operations accurately by operationId . `#546`_
 
 `1.4.0`_ - 2020-05-03
 ---------------------
@@ -1012,6 +1013,7 @@ Fixed
 .. _0.3.0: https://github.com/kiwicom/schemathesis/compare/v0.2.0...v0.3.0
 .. _0.2.0: https://github.com/kiwicom/schemathesis/compare/v0.1.0...v0.2.0
 
+.. _#546: https://github.com/kiwicom/schemathesis/issues/546
 .. _#542: https://github.com/kiwicom/schemathesis/issues/542
 .. _#539: https://github.com/kiwicom/schemathesis/issues/539
 .. _#537: https://github.com/kiwicom/schemathesis/issues/537
--- a/src/schemathesis/filters.py
+++ b/src/schemathesis/filters.py
@@ -26,3 +26,10 @@ def should_skip_by_tag(tags: Optional[List[str]], pattern: Optional[Filter]) -> 
         return True
     patterns = force_tuple(pattern)
     return not any(re.search(item, tag) for item in patterns for tag in tags)
+
+
+def should_skip_by_operation_id(operation_id: str, pattern: Optional[Filter]) -> bool:
+    if pattern is None:
+        return False
+    patterns = force_tuple(pattern)
+    return not any(re.search(item, operation_id) for item in patterns)
--- a/src/schemathesis/loaders.py
+++ b/src/schemathesis/loaders.py
@@ -25,6 +25,7 @@ def from_path(
     method: Optional[Filter] = None,
     endpoint: Optional[Filter] = None,
     tag: Optional[Filter] = None,
+    operation_id: Optional[Filter] = None,
     *,
     app: Any = None,
     validate_schema: bool = True,
@@ -38,6 +39,7 @@ def from_path(
             method=method,
             endpoint=endpoint,
             tag=tag,
+            operation_id=operation_id,
             app=app,
             validate_schema=validate_schema,
         )
@@ -49,6 +51,7 @@ def from_uri(
     method: Optional[Filter] = None,
     endpoint: Optional[Filter] = None,
     tag: Optional[Filter] = None,
+    operation_id: Optional[Filter] = None,
     *,
     app: Any = None,
     validate_schema: bool = True,
@@ -70,6 +73,7 @@ def from_uri(
         method=method,
         endpoint=endpoint,
         tag=tag,
+        operation_id=operation_id,
         app=app,
         validate_schema=validate_schema,
     )
@@ -82,6 +86,7 @@ def from_file(
     method: Optional[Filter] = None,
     endpoint: Optional[Filter] = None,
     tag: Optional[Filter] = None,
+    operation_id: Optional[Filter] = None,
     *,
     app: Any = None,
     validate_schema: bool = True,
@@ -99,6 +104,7 @@ def from_file(
         method=method,
         endpoint=endpoint,
         tag=tag,
+        operation_id=operation_id,
         app=app,
         validate_schema=validate_schema,
     )
@@ -111,6 +117,7 @@ def from_dict(
     method: Optional[Filter] = None,
     endpoint: Optional[Filter] = None,
     tag: Optional[Filter] = None,
+    operation_id: Optional[Filter] = None,
     *,
     app: Any = None,
     validate_schema: bool = True,
@@ -126,6 +133,7 @@ def from_dict(
             method=method,
             endpoint=endpoint,
             tag=tag,
+            operation_id=operation_id,
             app=app,
             validate_schema=validate_schema,
         )
@@ -139,6 +147,7 @@ def from_dict(
             method=method,
             endpoint=endpoint,
             tag=tag,
+            operation_id=operation_id,
             app=app,
             validate_schema=validate_schema,
         )
@@ -158,10 +167,18 @@ def from_pytest_fixture(
     method: Optional[Filter] = NOT_SET,
     endpoint: Optional[Filter] = NOT_SET,
     tag: Optional[Filter] = NOT_SET,
+    operation_id: Optional[Filter] = NOT_SET,
     validate_schema: bool = True,
 ) -> LazySchema:
     """Needed for a consistent library API."""
-    return LazySchema(fixture_name, method=method, endpoint=endpoint, tag=tag, validate_schema=validate_schema)
+    return LazySchema(
+        fixture_name,
+        method=method,
+        endpoint=endpoint,
+        tag=tag,
+        operation_id=operation_id,
+        validate_schema=validate_schema,
+    )
 
 
 def from_wsgi(
@@ -171,6 +188,7 @@ def from_wsgi(
     method: Optional[Filter] = None,
     endpoint: Optional[Filter] = None,
     tag: Optional[Filter] = None,
+    operation_id: Optional[Filter] = None,
     validate_schema: bool = True,
     **kwargs: Any,
 ) -> BaseSchema:
@@ -188,6 +206,7 @@ def from_wsgi(
         method=method,
         endpoint=endpoint,
         tag=tag,
+        operation_id=operation_id,
         app=app,
         validate_schema=validate_schema,
     )
@@ -206,6 +225,7 @@ def from_aiohttp(
     method: Optional[Filter] = None,
     endpoint: Optional[Filter] = None,
     tag: Optional[Filter] = None,
+    operation_id: Optional[Filter] = None,
     *,
     validate_schema: bool = True,
     **kwargs: Any,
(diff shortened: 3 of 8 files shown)


docs/changelog.rst
```diff
@@ -5,6 +5,7 @@ Changelog
 
 `Unreleased`_
 -------------
+- ``operationId`` argument can  get operations accurately by operationId (Only with OpenAPI 3.0+)
```
Stranger6667, line 8: Please, remove double space after "can" and add an empty line before this one. Also, I think that if we follow the general naming convention in the repo, then the argument name should be `operation_id`

docs/changelog.rst
```diff
@@ -18,6 +19,7 @@ Added
 - New ``before_process_path`` hook.
 - Third-party compatibility fixups mechanism. Currently there is one fixup for `FastAPI <https://github.com/tiangolo/fastapi>`_. `#503`_
 
+
```
Stranger6667, line 22: Please, remove this empty line

src/schemathesis/loaders.py
```diff
@@ -126,6 +133,7 @@ def from_dict(
             method=method,
             endpoint=endpoint,
             tag=tag,
+            operationId=None,  # Swagger does not support temporarily
```
Stranger6667, line 136: How Swagger is different in this matter? How much effort do we need to support it?

src/schemathesis/schemas.py
```diff
@@ -103,6 +105,9 @@ def resolver(self) -> jsonschema.RefResolver:
     def endpoints_count(self) -> int:
         return len(list(self.get_all_endpoints()))
 
+    def get_endpoint_by_operationId(self) -> Dict[str, Any]:
```
Stranger6667, line 108: I do think that we should distinguish two things:
- Filtering of endpoints based on `operationId` key;
- API for getting endpoints by `operationId`.

The latter is not needed for filtration, thus I'd avoid having it, filtration can be implemented in the same way as for `method` / `tag` or `endpoint`

src/schemathesis/schemas.py
```diff
@@ -232,8 +244,14 @@ def get_full_path(self, path: str) -> str:
 
     def get_all_endpoints(self) -> Generator[Endpoint, None, None]:
         try:
-            paths = self.raw_schema["paths"]  # pylint: disable=unsubscriptable-object
+            if self.operationId:
+                # use the operationId to uniquely identify an operation
+                paths = self.get_endpoint_by_operationId()
```
Stranger6667, line 249: The name seems misleading to me - I would expect that there will be a single endpoint returned instead of paths that I assume (from the variable name) may contain multiple values.

src/schemathesis/schemas.py
```diff
@@ -398,11 +416,39 @@ def get_content_types(self, endpoint: Endpoint, response: GenericResponse) -> Li
...
 
 
-class OpenApi30(SwaggerV20):  # pylint: disable=too-many-ancestors
+@attr.s()  # pylint: disable=too-many-ancestors
+class OpenApi30(SwaggerV20):
     nullable_name = "nullable"
     example_field = "example"
     operations = SwaggerV20.operations + ("trace",)
 
+    def __attrs_post_init__(self) -> None:
```
Stranger6667, line 425: I'd firmly avoid double iteration on paths. The first one will happen here and another one in `get_all_endpoints` if there is no filter by `operationId` (in this case there some traversing as well). I'd say that having extra runtime cost for comparing `operationId` value with the filter in `get_all_endpoints` has a lower performance impact and much easier to implement than the currently proposed approach.

I don't think that filtration by `operationId` should be implemented differently than other filters as well as having uniqueness check at all (it may be done but I'd say in a different pull request / issue)

test/test_filters.py
```diff
@@ -136,3 +136,54 @@ def test_b(request, case):
...
+        paths={
+            "/foo": {"get": {**parameters, "tags": ["foo", "baz"], "operationId": "foo_get"}},
+            "/bar": {"get": {**parameters, "tags": ["bar", "baz"], "operationId": "bar_get"}},
+        },
+        schema_name="simple_openapi.yaml",
+    )
+
+    result = testdir.runpytest("-v", "-s")
+    result.assert_outcomes(passed=1)
+    # Then only tests for this tag should be generated
```
Stranger6667, line 161: `tag` -> `operationId`

test/test_filters.py
```diff
@@ -136,3 +136,54 @@ def test_b(request, case):
...
+        """
+@schema.parametrize(operationId="bar_get")
+@settings(max_examples=1)
+def test_(request, case):
+    request.config.HYPOTHESIS_CASES += 1
+    assert case.path == "/v1/bar"
+    assert case.method == "GET"
+""",
+        paths={
+            "/foo": {"get": {**parameters, "tags": ["foo", "baz"], "operationId": "foo_get"}},
```
Stranger6667, line 153: Tags are redundant here and can be deleted to improve readability

test/test_filters.py
```diff
@@ -136,3 +136,54 @@ def test_b(request, case):
     result = testdir.runpytest("-v", "-s")
     result.assert_outcomes(passed=2)
     result.stdout.re_match_lines([r"Hypothesis calls: 2$"])
+
+
+def test_operationId_filter_with_openapi(testdir):
```
Stranger6667, line 141: Could you, please, rename `operationId` to `operation_id` to follow naming convention in other tests?

test/test_filters.py
```diff
@@ -136,3 +136,54 @@ def test_b(request, case):
...
+                "post": {**parameters, "tags": ["foo", "baz"], "operationId": "foo_post"},
+            },
+            "/bar": {"get": {**parameters, "tags": ["bar", "baz"], "operationId": "bar_get"}},
+        },
+        schema_name="simple_openapi.yaml",
+    )
+
+    result = testdir.runpytest("-v", "-s")
+    result.assert_outcomes(passed=2)
+    # Then only tests for this tag should be generated
```
Stranger6667, line 187: `tag` -> `operationId`

test/test_filters.py
```diff
@@ -136,3 +136,54 @@ def test_b(request, case):
...
+        """
+@schema.parametrize(operationId=["foo_get", "foo_post"])
+@settings(max_examples=1)
+def test_(request, case):
+    request.config.HYPOTHESIS_CASES += 1
+    assert case.path == "/v1/foo"
+""",
+        paths={
+            "/foo": {
+                "get": {**parameters, "tags": ["foo", "baz"], "operationId": "foo_get"},
```
Stranger6667, line 177: Tags are redundant here and can be deleted to improve readability

Stranger6667:
Also, we use conventional commits to format commit messages - it is the reason why the check failed.
https://www.conventionalcommits.org/en/

I'll update our contribution docs accordingly

src/schemathesis/schemas.py
Stranger6667: Probably I misunderstood your intentions here, this method supports other parts of the implementation

dongfangtianyu:
Hi, @Stranger6667 
Thank you for putting forward these, which will be very helpful to me.
I'll commit again it as soon as possible.

Cheers

dongfangtianyu:
I switch to new way， this like similar way to other filters.
The previous way is to consider that the map efficiency will be higher when there is a unique operationId, but obviously, at present, the efficiency is not the first consideration

about Swagger , will be resolved in next commit

docs/changelog.rst
dongfangtianyu: Yes, I modified it, and it conflicts with the latest file. I'll continue later

docs/changelog.rst
dongfangtianyu: Yes, I modified it,

src/schemathesis/schemas.py
dongfangtianyu: Deleted it,filtering will be run later in the code

src/schemathesis/schemas.py
dongfangtianyu: Sorry, I not get it, about `double iteration`. I will read it again after done this PR, will consult you then

test/test_filters.py
dongfangtianyu: Yes, I modified it

test/test_filters.py
dongfangtianyu: Yes, I modified it

src/schemathesis/filters.py
```diff
@@ -26,3 +26,10 @@ def should_skip_by_tag(tags: Optional[List[str]], pattern: Optional[Filter]) ->
         return True
     patterns = force_tuple(pattern)
     return not any(re.search(item, tag) for item in patterns for tag in tags)
+
+
+def should_skip_by_operation_id(operation_id: str, pattern: Optional[Filter]) -> bool:
+    if pattern is None:
+        return False
+    patterns = force_tuple(pattern)
+    return operation_id not in patterns
```
Stranger6667, line 35: probably we can also support regular expressions here, which will be the same implementation as in `should_skip_endpoint`

src/schemathesis/schemas.py
```diff
@@ -115,18 +117,20 @@ def get_all_tests(
...
+        operation_id: Optional[Filter] = NOT_SET,
         validate_schema: Union[bool, NotSet] = NOT_SET,
     ) -> Callable:
         """Mark a test function as a parametrized one."""
 
         def wrapper(func: GenericTest) -> GenericTest:
             HookDispatcher.add_dispatcher(func)
-            func._schemathesis_test = self.clone(func, method, endpoint, tag, validate_schema)  # type: ignore
+            func._schemathesis_test = self.clone(func, method, endpoint, tag, operation_id, validate_schema)  # type: ignore
+
```
Stranger6667, line 133: Could you, please, remove this extra line?

src/schemathesis/schemas.py
```diff
@@ -398,7 +409,8 @@ def get_content_types(self, endpoint: Endpoint, response: GenericResponse) -> Li
         return self.raw_schema.get("produces", [])
 
 
-class OpenApi30(SwaggerV20):  # pylint: disable=too-many-ancestors
+@attr.s()  # pylint: disable=too-many-ancestors
```
Stranger6667, line 412: I believe these two lines can be reverted since there is no need to have this decorator

src/schemathesis/schemas.py
```diff
@@ -234,6 +243,7 @@ def get_all_endpoints(self) -> Generator[Endpoint, None, None]:
         try:
             paths = self.raw_schema["paths"]  # pylint: disable=unsubscriptable-object
             context = HookContext()
+
```
Stranger6667, line 246: Could you, please, remove this empty line?

src/schemathesis/schemas.py
```diff
@@ -60,10 +60,12 @@ class BaseSchema(Mapping):
     method: Optional[Filter] = attr.ib(default=None)  # pragma: no mutate
     endpoint: Optional[Filter] = attr.ib(default=None)  # pragma: no mutate
     tag: Optional[Filter] = attr.ib(default=None)  # pragma: no mutate
+    operation_id: Optional[Filter] = attr.ib(default=None)  # pragma: no mutate
     app: Any = attr.ib(default=None)  # pragma: no mutate
     hooks: HookDispatcher = attr.ib(factory=lambda: HookDispatcher(scope=HookScope.SCHEMA))  # pragma: no mutate
     test_function: Optional[GenericTest] = attr.ib(default=None)  # pragma: no mutate
     validate_schema: bool = attr.ib(default=True)  # pragma: no mutate
+    schema_map: Union[Dict[str, Dict[str, Any]], None] = attr.ib(default=None)
```
Stranger6667, line 68: This attribute is not used anymore and can be removed

src/schemathesis/filters.py
dongfangtianyu: good idea! I haven't seen anything about Re in the documentation before, but this is really an exciting feature, maybe you can update the documentation? : P

dongfangtianyu:
Before， trying to use raw_schema directly,  encountered different  between swagger and opnapi, so  need to deal with them separately.

But now, with the new filtering method, there is no such problem. Swagger and OpenAPI will be supported at the same time,, so I change PR title : )

src/schemathesis/filters.py
Stranger6667: I'll add it to our todo list! Thank you for the suggestion. #500

Stranger6667:
Thanks once again @dongfangtianyu !
I'll merge it tomorrow

Issue #458, opened 2020-04-02
Add hooks to modify request before request is sent


barrett-schonefeld:
### Description

Schemathesis currently has hooks that allow users to extend the set of Schemathesis validation functions. I would like to be able to write hooks that allow me to alter the request before it is sent to the API. This will allow me to test specific behavior of the API. 

Example of one extended validation rule: 

    Accept header MUST be used to indicate allowable request methods in a 405 response.

To exercise this validation rule, we need to send a request with an operation that we know is not in the API definition for that endpoint. This idea is mentioned in #108, so this behavior may be included as part of the broader goal to generate invalid data in #65.

However, some of the rules are likely too specific to be part of the default request generation. For example:

    Requests including Accept header with only formats not supported by the server MUST be
    rejected with a 406 status code and appropriate error response model.

### Design

The validation hooks are simple and easy to use, and I think the request hooks should use the same design with different parameters. I think the should request hooks should take a request object and a case object (or any object that contains the API description for the endpoint associated with the request).

### Questions to Consider

- Does mutating an initially valid request cause problems?
- Do the extended validation rules still run when the status code is not in the API definition? 415 response, for example, may not be documented in the OpenAPI description.
- Which requests do we modify? We only want to modify a subset of the requests. Maybe we generate a valid request for each hook in addition to the default requests. Each additional valid request will be modified by a single hook.

Stranger6667:
Hi @barrett-schonefeld !
Thank you for the proposal, I like it a lot :) Here are my answers to the questions above:
1. No, it should not. It is actually constructed from a `Case` instance [here](https://github.com/kiwicom/schemathesis/blob/master/src/schemathesis/models.py#L123). Maybe it will be enough for a hook to take only a `Case` instance then. What do you think? E.g. a hook might be called right after `case` creation
2. Checks are called on all responses ([code](https://github.com/kiwicom/schemathesis/blob/master/src/schemathesis/runner/impl/core.py#L130)). Most probably `status_code_conformance` will fail if any not documented response code will appear (and there is no `default` included in `responses`)
3. Hm, it feels like filtration should be on the hook side, so the user may decide what to modify. 

Re: generating extra requests. I am not sure, actually, need to re-check the architecture and think more about it. E.g. how to report the results from such modified requests

barrett-schonefeld:
Great, I have a couple follow-ups!

[...]

PR #564, opened 2020-05-05
WIP: feat: Add generate_test_per_endpoint hook

barrett-schonefeld:
Related issue: #458 

This initial PR is a proof of concept and is ready for discussion.

This solution adds a `before_one` hook. This hook adds a test to the test suite for each `before_one` hook defined. Each `before_one` hook receives one of the strategies from one of the additional tests. The hook may update this strategy before a request is sent to the server. Since each strategy produces data that is valid against the schema, it allows the user to alter an otherwise valid strategy to produce requests that target specific behavior of the server.

Additional Questions and Needs:
- Tests
- Does this implementation duplicate requests generated from an example (i.e. `--hypothesis-phases=explicit`)? Because an example request is (in my experience) the best representation of a valid request, I would like the duplicated tests to come from examples if they exist and randomly generated, valid data otherwise.
- Is the `GLOBAL` scope appropriate for this hook?
- Is it OK that we have to get the `before_one` hooks in two places (threadpool.py and schemas.py)?

### DIFF
--- a/src/schemathesis/_hypothesis.py
+++ b/src/schemathesis/_hypothesis.py
@@ -3,8 +3,9 @@ import asyncio
 import inspect
 import re
 from base64 import b64encode
+from copy import deepcopy
 from functools import partial
-from typing import Any, Callable, Dict, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Union
 from urllib.parse import quote_plus
 
 import hypothesis
@@ -15,7 +16,7 @@ from requests.auth import _basic_auth_str
 from . import utils
 from ._compat import handle_warnings
 from .exceptions import InvalidSchema
-from .hooks import GLOBAL_HOOK_DISPATCHER, HookContext, HookDispatcher
+from .hooks import GLOBAL_HOOK_DISPATCHER, HookContext, HookDispatcher, get_all_by_name
 from .models import Case, Endpoint
 from .types import Hook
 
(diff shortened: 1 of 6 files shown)


Stranger6667:
Hi @barrett-schonefeld 

Sorry for not getting back to you. Thank you for your contribution, I'll try to find a timeslot to review your PR in the next few days :)

src/schemathesis/hooks.py
```diff
@@ -222,6 +222,11 @@ def before_load_schema(context: HookContext, raw_schema: Dict[str, Any]) -> None
     """Called before schema instance is created."""
 
 
+@HookDispatcher.register_spec([HookScope.GLOBAL])
+def before_one(strategy: st.SearchStrategy) -> st.SearchStrategy:
+    """Creates an additional test and is called before one valid request."""
```
Stranger6667, line 227: The hook name sounds a bit confusing to me since it's behavior actually implies generating new tests. May be something `generate_tests`? or something like that, maybe we can reflect also that these tests are generated per endpoint

src/schemathesis/_hypothesis.py
```diff
@@ -24,11 +24,17 @@
 
 
 def create_test(
-    endpoint: Endpoint, test: Callable, settings: Optional[hypothesis.settings] = None, seed: Optional[int] = None
+    endpoint: Endpoint,
+    test: Callable,
+    settings: Optional[hypothesis.settings] = None,
+    seed: Optional[int] = None,
+    before_one_hook: Optional[Callable] = None,
```
Stranger6667, line 31: I am thinking if we can make `create_test` a generator, so it always generates tests, in this way we won't have to iterate one extra time on the callee side as it is now in `_get_tests_with_hooks` and `get_all_tests`. Also, we will not need to pass `before_one_hook` argument here

Maybe, in this case, we can also rename `create_test` to something more appropriate

src/schemathesis/_hypothesis.py
```diff
@@ -24,11 +24,17 @@
...
 ) -> Callable:
     """Create a Hypothesis test."""
     hook_dispatcher = getattr(test, "_schemathesis_hooks", None)
     strategy = endpoint.as_strategy(hooks=hook_dispatcher)
+    if before_one_hook:
```
Stranger6667, line 36: I believe, that we can use schema- and test-local hooks as well here if this function will be a generator

src/schemathesis/hooks.py
```diff
@@ -222,6 +222,11 @@ def before_load_schema(context: HookContext, raw_schema: Dict[str, Any]) -> None
     """Called before schema instance is created."""
 
 
+@HookDispatcher.register_spec([HookScope.GLOBAL])
```
Stranger6667, line 225: I think that it might be done on all scopes

Stranger6667:
Regarding your questions:

> Tests

I think that we need to test it in CLI & in-code cases. Usually, I do such kind of tests with the `testdir` fixture. There are some examples in `test/cli/test_commands.py`. Let me know if I can provide more information on tests

> Does this implementation duplicate requests generated from an example (i.e. --hypothesis-phases=explicit)? Because an example request is (in my experience) the best representation of a valid request, I would like the duplicated tests to come from examples if they exist and randomly generated, valid data otherwise.

Examples are added a bit later in `create_test`, so examples will be duplicated since this function creates separate hypothesis test functions and each will have its own `@example` decorator

> Is the GLOBAL scope appropriate for this hook?

I believe it could be on all scopes. It might be easier to implement if hooks will be extracted inside `create_test` (there are local hook dispatcher and `endpoint.schema.hooks` so it is easier to get in a single place)

> Is it OK that we have to get the before_one hooks in two places (threadpool.py and schemas.py)?

Probably it is better to isolate inside `create_test`. At least now I think that it might be easier and more flexible if it will be implemented as a generator (as I described in comments)

src/schemathesis/hooks.py
```diff
@@ -222,6 +222,11 @@ def before_load_schema(context: HookContext, raw_schema: Dict[str, Any]) -> None
...
+def before_one(strategy: st.SearchStrategy) -> st.SearchStrategy:
```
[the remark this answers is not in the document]
barrett-schonefeld: Added `context` parameter. I made `context` the second parameter to match the format of `before_generate*(strategy: st.SearchStrategy, context: HookContext)`.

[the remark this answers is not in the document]
Stranger6667: Oh ... I completely forgot to update the docs. In the actual code, it is passed as the first argument :( I will update the docs

src/schemathesis/hooks.py
barrett-schonefeld: I updated the name to `generate_test_per_operation`. This is verbose but descriptive, so please let me know if we should keep refining the name.

src/schemathesis/hooks.py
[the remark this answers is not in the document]
barrett-schonefeld: That's all right! I will update the signatures in this PR (if we want the fix merged into master faster in another PR, I'll just consider it an effort to avoid merge conflicts).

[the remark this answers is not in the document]
Stranger6667: Cool :) Fixed in master in e846790fdf988d9ac836c948d9d6599329077c72

src/schemathesis/_hypothesis.py
barrett-schonefeld: I made `create_test` a generator. 

However, because `create_test` may throw an `InvalidSchema` exception, all the generated values must be consumed and put in a list to see if an `InvalidSchema` exception is thrown. I also wrap the `InvalidSchema` exception in a list, so we may still iterate on the return value from `make_test_or_exception`.

I will push the changes, so you can see the code (most likely today). I'm wondering if it makes more sense for `create_test` to return a list instead of a generator.

Stranger6667:
Looks great! Could you, please, rebase?

I can do some more review tomorrow

barrett-schonefeld:
Sounds great! Will rebase and push latest updates.

Still need to:
- Test my implementation.
- Know if the hooks will modify requests generated from OpenAPI examples. My intuition is that the current implementation will not modify requests generated from OpenAPI examples, but I need to explore this more thoroughly.

barrett-schonefeld:
@Stranger6667 how a strategy converted to a `Case`?

It seems `SearchStrategy` and `Case` are used somewhat interchangeably. For example, [this get_example function](https://github.com/kiwicom/schemathesis/blob/master/src/schemathesis/_hypothesis.py#L75) returns a `Case`. The [return statement](https://github.com/kiwicom/schemathesis/blob/master/src/schemathesis/_hypothesis.py#L88), however, uses a function that returns a `SearchStrategy`.

I don't see any explicit relationships between them in code (neither object is a descendant of the other). Could you explain how they are related and possibly point to code that shows how the conversion is done?

Stranger6667:
@barrett-schonefeld 

> Is there a particular piece of code responsible for getting Case objects generated from Hypothesis.core.given?

This part is [here](https://github.com/kiwicom/schemathesis/blob/master/src/schemathesis/_hypothesis.py#L199). The idea is to use `st.builds` that will create instances of `Case` during test runs. But, for examples, we need to have instances before a test run to add them via `hypothesis.example` decorator. Often examples are incomplete and we take static parts from `example` fields in the schema and generate the rest by using `.example()` call on a strategy. `example` is often is not representative and is not recommended to use (`given` is preferred), but we don't care about these disadvantages since we need to fill gaps in a `Case` instance that will be used. Also, it is possible, that the schema has all data and in this case, generation will be a no-op (but actually slower than it could be - we can just create an instance straight away in this case). More info on how `builds` works is [here](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.builds).

> I have not found code where Case.call is used to make a request.

There are two places:
- HTTP is in the `network_test` function [here](https://github.com/kiwicom/schemathesis/blob/master/src/schemathesis/runner/impl/core.py#L165)
- WSGI is in the `wsgi_test` function [here](https://github.com/kiwicom/schemathesis/blob/master/src/schemathesis/runner/impl/core.py#L202)

> For this PR, would it be better for the generate_test_per_endpoint hooks to accept a Case or a SearchStrategy?

I thought that it aims to generate an additional test in a sense of an extra endpoint :( With the current approach, it will be generated per endpoint (which means also an extra line in the output). However, it could be done this way by creating a strategy with one possible output, but if it should be one test (in a sense of request) then it is better to have a hook somewhere below in the runner, for example in `network_test`. 
It seems to me that both approaches could be useful in different circumstances. The approach with a hook inside of `network_test` / `wsgi_test` should be simpler to implement, but we'll need to decide how to treat the statistic - we can just call `run_checks` & `store_requests_response` on the generated extra response (not sure about `run_targets`).
Having all of this I think that the latter approach can be represented as a hook that accepts `Case` instance. And it looks like that the goals can be fulfilled with this approach

src/schemathesis/_hypothesis.py
Stranger6667: I think that this generator will be consumed anyway in a few next steps, since `collect` inside the pytest plugin will consume it before proceeding, so it will be done during the collection phase anyway. Currently, if there are e.g. 5 generated tests and an error happens, then all 5 will be replaced with one lambda. I am not sure, but there could be 4 valid tests that can be executed - maybe it would be better to check each instance during the iteration, and if it is failing then generate that lambda with `pytest.fail`. What do you think?

barrett-schonefeld:
Closing this PR in favor of #590.

PR #581, opened 2020-05-10
feat: OpenAPI links support

Stranger6667:
Very early PoC PR to keep track of TODOs & comment on code

TODO:

- [x] Quite a lot of small things in comments
- [x] Threaded runners support
- [x] WSGI support
- [x] Fix coverage
- [x] Docs
- [x] More docstrings
- [x] complete runtime expressions support. https://swagger.io/docs/specification/links/#runtime-expressions 
- [x] Support for location prefixes in links when two or more parameters have the same name (`query.id`, `body.id`, etc)
- [x] What if the parsed expression is not a string?
- [x] `$response` variable support
- [x] JSON pointer support
- [x] Keep the same type inside runtime expressions
- [x] Hypothesis tests for runtime expressions
- [x] Entire body support in runtime expressions. How it should be done? JSON?
- [x] Error handling for runtime expressions (including unbalanced brackets)
- [x] Handle recursion
- [x] Handle CLI percentage properly
- [x] Deduplicate data samples
- [x] Add a distinguishable CLI name to the test generated by the `feedback` instance
- [x] Fix not needed newline in CLI output for threaded execution
- [x] Store endpoints by `operationId`
- [x] Resolve `operationRef` - local & external
- [x] Resolve links behind references
- [x] CLI option to enable links. E.g. `--stateful=links`
- [x] Proper support for `requestBody`
- [x] handle empty `parameters`
- [x] gather common parameters when resolving operation by `operationRef`
- [x] test for resolving external reference in `operationRef`
- [x] Configurable recursion limit + test
- [x] non JSON payload in `request.body`

In the future:
- `server` keyword overriding

Maybe the CLI output could look like this:

```
POST /api/users/ . 
  -> GET /api/users/{user_id} .
  -> PUT /api/users/{user_id} .
       -> DELETE /api/users/{user_id} .
```

So it is clear what is the dependency tree

### DIFF
--- a/src/schemathesis/runner/impl/core.py
+++ b/src/schemathesis/runner/impl/core.py
@@ -1,7 +1,7 @@
 import logging
 import time
 from contextlib import contextmanager
-from typing import Any, Callable, Dict, Generator, Iterable, List, Optional, Union
+from typing import Any, Callable, Dict, Generator, Iterable, List, Optional, Tuple, Union
 
 import attr
 import hypothesis
@@ -9,22 +9,67 @@ import requests
 from _pytest.logging import LogCaptureHandler, catching_logs
 from requests.auth import HTTPDigestAuth, _basic_auth_str
 
+from ..._hypothesis import make_test_or_exception
 from ...constants import USER_AGENT
 from ...exceptions import InvalidSchema, get_grouped_exception
 from ...models import Case, CheckFunction, Endpoint, Status, TestResult, TestResultSet
 from ...runner import events
 from ...schemas import BaseSchema
+from ...stateful import ParsedData, StatefulTest
 from ...types import RawAuth
 from ...utils import GenericResponse, capture_hypothesis_output
 from ..targeted import Target
 
 DEFAULT_DEADLINE = 500  # pragma: no mutate
+DEFAULT_STATEFUL_RECURSION_LIMIT = 5  # pragma: no mutate
 
 
 def get_hypothesis_settings(hypothesis_options: Dict[str, Any]) -> hypothesis.settings:
     # Default settings, used as a parent settings object below
     hypothesis_options.setdefault("deadline", DEFAULT_DEADLINE)
     return hypothesis.settings(**hypothesis_options)
+
+
+@attr.s(slots=True)
+class StatefulData:
+    """Storage for data that will be used in later tests."""
+
+    stateful_test: StatefulTest = attr.ib()
+    container: List[ParsedData] = attr.ib(factory=list)
+
+    def make_endpoint(self) -> Endpoint:
+        return self.stateful_test.make_endpoint(self.container)
+
+    def store(self, case: Case, response: GenericResponse) -> None:
+        """Parse and store data for a stateful test."""
+        parsed = self.stateful_test.parse(case, response)
+        self.container.append(parsed)
+
+
+@attr.s(slots=True)
+class Feedback:
+    """Handler for feedback from tests.
+
+    Provides a way to control runner's behavior from tests.
+    """
+
+    stateful: Optional[str] = attr.ib()
+    endpoint: Endpoint = attr.ib()
+    stateful_tests: Dict[str, StatefulData] = attr.ib(factory=dict)
+
+    def add_test_case(self, case: Case, response: GenericResponse) -> None:
+        """Store test data to reuse it in the future additional tests."""
+        for stateful_test in case.endpoint.get_stateful_tests(response, self.stateful):
+            data = self.stateful_tests.setdefault(stateful_test.name, StatefulData(stateful_test))
+            data.store(case, response)
+
+    def get_stateful_tests(
+        self, test: Callable, settings: hypothesis.settings, seed: Optional[int]
+    ) -> Generator[Tuple[Endpoint, Union[Callable, InvalidSchema]], None, None]:
+        """Generate additional tests that use data from the previous ones."""
+        for data in self.stateful_tests.values():
+            endpoint = data.make_endpoint()
+            yield endpoint, make_test_or_exception(endpoint, test, settings, seed)
 
 
 # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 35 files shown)


src/schemathesis/specs/openapi/links.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def get_links(response: GenericResponse, endpoint: Endpoint, field: str) -> List[Link]:
+    """Get `x-links` / `links` definitions from the schema."""
+    responses = endpoint.definition.raw["responses"]
+    if str(response.status_code) in responses:
+        response_definition = responses[str(response.status_code)]
+    else:
+        response_definition = responses.get("default", {})
+    links = response_definition.get(field, {})
```
Stranger6667, line 32: Links could be behind references

src/schemathesis/specs/openapi/links.py
```diff
@@ -0,0 +1,33 @@
...
+    name: str = attr.ib()
+    endpoint: Endpoint = attr.ib()
+    parameters: Dict[str, Any] = attr.ib()
+
+    @classmethod
+    def from_definition(cls, name: str, definition: Dict[str, Dict[str, Any]], source_endpoint: Endpoint) -> "Link":
+        endpoint = None
+        for endpoint in source_endpoint.schema.get_all_endpoints():
+            # it could be better to have O(1) access via a mapping by operationId or operationRef
+            if endpoint.definition.raw.get("operationId") == definition.get("operationId"):
```
Stranger6667, line 20: I think that it will be much better to build a structure to access by a key instead of iterating each time

src/schemathesis/specs/openapi/expressions/lexer.py
```diff
@@ -0,0 +1,104 @@
...
+    @property
+    def is_variable(self) -> bool:
+        return self.type_ == TokenType.VARIABLE
+
+    @property
+    def is_dot(self) -> bool:
+        return self.type_ == TokenType.DOT
+
+
+def tokenize(expression: str) -> Generator[LexToken, None, None]:
```
Stranger6667, line 59: Probably tokenization result could be cached to avoid re-tokenizing the same expressions

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -20,13 +21,17 @@
...
     nullable_name: str
+    links_field: str
     operations: Tuple[str, ...]
     security: BaseSecurityProcessor
 
     @property  # pragma: no mutate
     def spec_version(self) -> str:
         raise NotImplementedError
 
+    def get_additional_tests(self, response: GenericResponse, endpoint: Endpoint) -> List:
```
Stranger6667, line 32: It could be a non-generic type. Maybe some base class for additional test

test/apps/utils.py
```diff
@@ -152,6 +155,38 @@ def make_schema(endpoints: Tuple[str, ...]) -> Dict:
...
+                            "type": "object",
+                            "properties": {"username": {"type": "string", "minLength": 3}},
+                            "required": ["username"],
+                        },
+                    }
+                ],
+                "responses": {
+                    "201": {
+                        "description": "OK",
+                        "x-links": {
```
Stranger6667, line 175: Since there are no links in Open API 2 we need to handle `requestBody` carefully. Either we need to name it `body` and handle like other parameters or still use `requestBody`, however, it is not clear for me how to specify multiple parameters for `requestBody`

src/schemathesis/specs/openapi/expressions/parser.py
```diff
@@ -0,0 +1,90 @@
...
+
+def _parse(expr: str) -> Generator[nodes.Node, None, None]:
+    # pylint: disable=too-many-branches
+    tokens = lexer.tokenize(expr)
+    brackets_stack: List[str] = []
+    for token in tokens:
+        if token.is_string:
+            yield nodes.String(token.value)
+        elif token.is_variable:
+            if token.value == "$url":
```
Stranger6667, line 22: `$url` and other strings could be in a separate enum, so they are defined in a single place

src/schemathesis/runner/impl/core.py
```diff
@@ -27,6 +28,35 @@ def get_hypothesis_settings(hypothesis_options: Dict[str, Any]) -> hypothesis.se
...
+
+    Provides a way to control runner's behavior from tests.
+    """
+
+    endpoint: Endpoint = attr.ib()
+    data: Dict = attr.ib(factory=dict)
+
+    def add(self, case: Case, response: GenericResponse) -> None:
+        # Init all possible links after init to avoid iteration over them every time
+        for link in case.endpoint.get_links(response):
```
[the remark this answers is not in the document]
Stranger6667: After adding `get_endpoint_by_operation_id` it doesn't take long anymore

src/schemathesis/runner/impl/core.py
```diff
@@ -27,6 +28,35 @@ def get_hypothesis_settings(hypothesis_options: Dict[str, Any]) -> hypothesis.se
...
+            data = self.data.setdefault(link.name, {"link": link, "data": []})
+            data["data"].append({"user_id": response.json().get("id")})
+
+    def process(
+        self, test: Callable, settings: hypothesis.settings, seed: Optional[int]
+    ) -> Generator[Tuple[Endpoint, Union[Callable, InvalidSchema]], None, None]:
+        if self.data:
+            for defs in self.data.values():
+                link = defs["link"]
+                # Instead of making all relevant requests we use `enum` to randomize them and enable shrinking
```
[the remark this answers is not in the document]
Stranger6667: `oneOf` is about the same, but it is probably should go to the `hypothesis-jsonschema` side

test/apps/utils.py
Stranger6667: Noted in the TODO list

Stranger6667:
Occasionally failing tests happen because of this - https://github.com/HypothesisWorks/hypothesis/issues/2433

PR #586, opened 2020-05-12
fix: on CLI mode , `hooks` file existence in cwd ,but ModuleNotFoundE…

dongfangtianyu:
## To Reproduce

Hooks definition in `myproject/hooks.py`

```bash
|-- myproject
|   |-- __init__.py
|   `-- hooks.py

````

than run schemathesis

```bash
schemathesis --pre-run myproject.hooks run http://127.0.0.1/openapi.yaml
```

Get an error

```python
ModuleNotFoundError: No module named 'myproject'
```
## Cause
on CLI mode, cwd not in `sys.path`

-----

The original test case [/test/cli/test_commands.py#L771](https://github.com/kiwicom/schemathesis/blob/master/test/cli/test_commands.py#L771) ,  
Test passed  may be because called `make_importable` and `module._ensuresyspath` changed `sys.path`

So I add  new testcase  [test/cli/test_commands.py#L784](https://github.com/dongfangtianyu/schemathesis/blob/master/test/cli/test_commands.py#L784)

### DIFF
--- a/src/schemathesis/cli/__init__.py
+++ b/src/schemathesis/cli/__init__.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import traceback
 from enum import Enum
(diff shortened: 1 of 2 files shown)


Stranger6667:
Good point! I actually run it locally with `PYTHONPATH=$(cwd)` which basically does the trick, but I didn't think about adding it to the codebase itself

dongfangtianyu:
Greart!
On Windows, can use  ` set PYTHONPATH=%cd%`

This may need to be added to the document?

Stranger6667:
I think it is OK :) Thank you!

PR #590, opened 2020-05-15
feat: add_case hooks

barrett-schonefeld:
Related issue: #458 

Updates:
- Adds `add_case` hook. For each `add_case` hook and for each endpoint, an additional, duplicate test is created, and the associated Case object is passed to hook before the request is sent to the server.

Purpose:
- Used to create targeted requests. The duplicate Case objects may be altered in specific ways to target specific behavior in the API. User may test, for example, how the API behavior is affected when the rest of the request stays the same but the `Content-Type` header is set to `application/json; charset=utf8`.

Tests:
- Adds a test to ensure that one additional Case object is created for each endpoint.
- Add test to ensure that mutation of a Case in an `add_case` hook does not affect other Cases.
- Adds a test to ensure the output is formatted correctly when the additional cases fail a check.
- Adds a test to ensure `Case.partial_deepcopy` deep copies expected fields.

Related PR: #564 

The related PR also creates additional tests and passes the associated strategy to the user for updates. This `add_case` hook is the behavior I meant to implement initially, but @Stranger6667 you mentioned that #564 may provide some good utility. I'm happy to finish and clean up #564 as well if you think it would be a good addition to the project.

Related issue: #458

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -11,6 +11,7 @@ Added
 
 - Support for YAML files in references via HTTP & HTTP schemas. `#600`_
 - Stateful testing support via ``Open API links`` syntax. `#548`_
+- New ``add_case`` hook. `#458`_
 
 Changed
 ~~~~~~~
@@ -1124,6 +1125,7 @@ Fixed
 .. _#468: https://github.com/kiwicom/schemathesis/issues/468
 .. _#463: https://github.com/kiwicom/schemathesis/issues/463
 .. _#461: https://github.com/kiwicom/schemathesis/issues/461
+.. _#458: https://github.com/kiwicom/schemathesis/issues/458
 .. _#457: https://github.com/kiwicom/schemathesis/issues/457
 .. _#451: https://github.com/kiwicom/schemathesis/issues/451
 .. _#450: https://github.com/kiwicom/schemathesis/issues/450
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -188,3 +188,19 @@ With this simple handler only ``Done!`` will be displayed at the end of the test
 - Store logs in a custom format
 - Change the output visual style
 - Display additional information in the output
+
+``add_case``
+~~~~~~~~~~~~
+
+For each ``add_case`` hook and for each endpoint, we create an additional, duplicate test case. We pass the Case object from the duplicate test to the ``add_case`` hook.
+The user may change the Case object (and therefore the request data) before the request is sent to the server. The ``add_case`` allows the user to target specific
+behavior in the API by changing specific details of the duplicate request.
+
+.. code:: python
+
+    def add_case(context: HookContext, case: Case) -> Case:
+        case.headers["Content-Type"] = "application/json"
+        return case
+
+Note: A partial deep copy of the ``Case`` object is passed to each ``add_case`` hook. ``Case.endpoint.app`` is a reference to the original ``app``, 
+and ``Case.endpoint.schema`` is a shallow copy, so changes to these fields will be reflected in other tests.
--- a/src/schemathesis/models.py
+++ b/src/schemathesis/models.py
@@ -4,6 +4,7 @@ import datetime
 import http
 from collections import Counter
 from contextlib import contextmanager
+from copy import deepcopy
 from enum import IntEnum
 from logging import LogRecord
 from typing import TYPE_CHECKING, Any, Callable, Dict, Generator, Iterator, List, Optional, Sequence, Tuple, Union, cast
@@ -198,6 +199,17 @@ class Case:
         prepared = requests.Session().prepare_request(request)  # type: ignore
         return prepared.url
 
+    def partial_deepcopy(self) -> "Case":
+        return self.__class__(
+            endpoint=self.endpoint.partial_deepcopy(),
+            path_parameters=deepcopy(self.path_parameters),
+            headers=deepcopy(self.headers),
+            cookies=deepcopy(self.cookies),
+            query=deepcopy(self.query),
+            body=deepcopy(self.body),
+            form_data=deepcopy(self.form_data),
+        )
+
 
 def is_multipart(item: Optional[Body]) -> bool:
     """A poor detection if the body should be a multipart request.
@@ -274,6 +286,22 @@ class Endpoint:
 
     def get_stateful_tests(self, response: GenericResponse, stateful: Optional[str]) -> Sequence["StatefulTest"]:
         return self.schema.get_stateful_tests(response, self, stateful)
+
+    def partial_deepcopy(self) -> "Endpoint":
+        return self.__class__(
+            path=self.path,  # string, immutable
+            method=self.method,  # string, immutable
+            definition=deepcopy(self.definition),
+            schema=self.schema.clone(),  # shallow copy
+            app=self.app,  # not deepcopyable
+            base_url=self.base_url,  # string, immutable
+            path_parameters=deepcopy(self.path_parameters),
+            headers=deepcopy(self.path_parameters),
+            cookies=deepcopy(self.cookies),
+            query=deepcopy(self.query),
+            body=deepcopy(self.body),
+            form_data=deepcopy(self.form_data),
+        )
 
 
 class Status(IntEnum):
--- a/src/schemathesis/runner/impl/core.py
+++ b/src/schemathesis/runner/impl/core.py
@@ -12,6 +12,7 @@ from requests.auth import HTTPDigestAuth, _basic_auth_str
 from ..._hypothesis import make_test_or_exception
 from ...constants import USER_AGENT
 from ...exceptions import InvalidSchema, get_grouped_exception
+from ...hooks import HookContext, get_all_by_name
 from ...models import Case, CheckFunction, Endpoint, Status, TestResult, TestResultSet
 from ...runner import events
 from ...schemas import BaseSchema
@@ -218,6 +219,13 @@ def run_targets(targets: Iterable[Target], elapsed: float) -> None:
             hypothesis.target(elapsed, label="response_time")
 
 
+def add_cases(case: Case, test: Callable, *args: Any) -> None:
+    context = HookContext(case.endpoint)
+    for case_hook in get_all_by_name("add_case"):
+        _case = case_hook(context, case.partial_deepcopy())
+        test(_case, *args)
+
+
 def network_test(
     case: Case,
     checks: Iterable[CheckFunction],
@@ -234,6 +242,22 @@ def network_test(
     headers = headers or {}
     headers.setdefault("User-Agent", USER_AGENT)
     timeout = prepare_timeout(request_timeout)
+    _network_test(case, checks, targets, result, session, timeout, store_interactions, headers, feedback)
+    add_cases(case, _network_test, checks, targets, result, session, timeout, store_interactions, headers, feedback)
+
+
+def _network_test(
+    case: Case,
+    checks: Iterable[CheckFunction],
+    targets: Iterable[Target],
+    result: TestResult,
+    session: requests.Session,
+    timeout: Optional[float],
+    store_interactions: bool,
+    headers: Optional[Dict[str, Any]],
+    feedback: Feedback,
+) -> None:
+    # pylint: disable=too-many-arguments
     response = case.call(session=session, headers=headers, timeout=timeout)
     run_targets(targets, response.elapsed.total_seconds())
     if store_interactions:
@@ -271,6 +295,20 @@ def wsgi_test(
 ) -> None:
     # pylint: disable=too-many-arguments
     headers = _prepare_wsgi_headers(headers, auth, auth_type)
+    _wsgi_test(case, checks, targets, result, headers, store_interactions, feedback)
+    add_cases(case, _wsgi_test, checks, targets, result, headers, store_interactions, feedback)
+
+
+def _wsgi_test(
+    case: Case,
+    checks: Iterable[CheckFunction],
+    targets: Iterable[Target],
+    result: TestResult,
+    headers: Dict[str, Any],
+    store_interactions: bool,
+    feedback: Feedback,
+) -> None:
+    # pylint: disable=too-many-arguments
     with catching_logs(LogCaptureHandler(), level=logging.DEBUG) as recorded:
         start = time.monotonic()
         response = case.call_wsgi(headers=headers)
--- a/test/cli/test_commands.py
+++ b/test/cli/test_commands.py
@@ -18,6 +18,7 @@ from schemathesis import Case, fixups
 from schemathesis._compat import metadata
 from schemathesis.checks import ALL_CHECKS
 from schemathesis.cli import reset_checks
+from schemathesis.hooks import unregister_all
 from schemathesis.loaders import from_uri
 from schemathesis.models import Endpoint
 from schemathesis.runner import DEFAULT_CHECKS, DEFAULT_TARGETS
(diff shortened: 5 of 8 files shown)


docs/changelog.rst
```diff
@@ -14,6 +14,11 @@ Fixed
 
 - Partially generated explicit examples are always valid and can be used in requests. `#582`_
 
+Added
```
Stranger6667, line 17: Could you, please, move this block above under the `Unreleased` section?

Also, one empty line should be added after the `~~~~~` line and there should be a link added to the bottom of the file:

```_#458: https://github.com/kiwicom/schemathesis/issues/458```

File: docs/customization.rst [hunk 2]
```diff
@@ -145,6 +145,23 @@ With this hook you can add additional test cases that will be executed in Hypoth
             Case(endpoint=context.endpoint, query={"foo": "bar"})
         )
 
+``add_case``
+~~~~~~~~~~~~~~~~~~~~~~~
```
Stranger6667, line 149: Could you, please,  reduce this line so it matches the length of the section name? it should be 12 symbols (`len("``add_case``"`)

docs/customization.rst
```diff
@@ -145,6 +145,23 @@ With this hook you can add additional test cases that will be executed in Hypoth
...
+
+For each add_case hook and for each endpoint, we create an additional, duplicate test case. We pass the Case object from the duplicate test to the add_case hook.
```
Stranger6667, line 151: Could you, please, wrap each `add_case` name in double-quotes so it is highlighted:

```
``add_case``
```

docs/customization.rst
```diff
@@ -145,6 +145,23 @@ With this hook you can add additional test cases that will be executed in Hypoth
...
+The user may change the Case object (and therefore the request data) before the request is sent to the server. The add_case allows the user to target specific
+behavior in the API by changing specific details of the duplicate request.
+
+.. code:: python
+
+    def add_case(
```
Stranger6667, line 157: Even though we have access to `endpoint` via case I think that the first argument should be hook context (at least for the forward-compatibility). What do you think?

Also, the function signature will be quite short - probably it can fit a single line

src/schemathesis/runner/impl/solo.py
```diff
@@ -17,37 +18,41 @@ def _execute(self, results: TestResultSet) -> Generator[events.ExecutionEvent, N
         auth = get_requests_auth(self.auth, self.auth_type)
         with get_session(auth) as session:
             for endpoint, test in self.schema.get_all_tests(network_test, self.hypothesis_settings, self.seed):
+                for case_hook in get_add_case_hooks():
```
Stranger6667, line 21: If `get_add_case_hooks` is here, then it will double the number of hypothesis tests. I am not sure that it is what you described earlier. Each test will be on a separate CLI line with the same name. I thought that you wanted a single extra request that is executed after the original one. All these requests will be executed after N hypothesis tests related to the previous endpoint. What do you think? Is it the kind of behavior you wanted?

src/schemathesis/runner/impl/core.py
```diff
@@ -156,9 +158,12 @@ def network_test(
     request_timeout: Optional[int],
     store_interactions: bool,
     headers: Optional[Dict[str, Any]],
+    case_hook: Optional[Callable[[Case], Case]],
 ) -> None:
     """A single test body that will be executed against the target."""
     # pylint: disable=too-many-arguments
+    if case_hook:
+        case = case_hook(deepcopy(case))
```
Stranger6667, line 166: There is no need for `deepcopy` since it is a unique object that is created for this test

test/cli/test_commands.py
```diff
@@ -798,6 +799,39 @@ def test_pre_run_hook_module_not_found(testdir, cli):
...
+def test_add_case(testdir, cli, schema_url):
+    module = testdir.make_importable_pyfile(
+        hook="""
+            import schemathesis
+            import click
+
+            @schemathesis.hooks.register
+            def add_case(case):
+                if not case.headers:
+                    case.headers = dict()
```
Stranger6667, line 811: As for small stylistic preference and small performance win I prefer having dicts as literals if it is possible - `{}`

test/cli/test_commands.py
```diff
@@ -798,6 +799,39 @@ def test_pre_run_hook_module_not_found(testdir, cli):
...
+    result = cli.main(
+        "--pre-run", module.purebasename, "run", "-c", "add_case_check", schema_url, "--hypothesis-max-examples=1"
+    )
+
+    assert result.exit_code == ExitCode.OK
+    # One additional case created for two endpoints - /api/failure and /api/success.
+    assert result.stdout.count("The case was added!") == 2
+    # resetting hooks and checks to ensure other tests start with clean slate
+    unregister_all()
+    reset_checks()
```
Stranger6667, line 832: To ensure that these lines are executed it would be more reliable to put them into a separate fixture:

```
@pytest.fixture
def reset_hooks():
    yield
    unregister_all()
    reset_checks()
```

And then add it as a dependency for this test:

```
@pytest.mark.usefixtures("reset_hooks")
def test_add_case(testdir, cli, schema_url):
    ...
```

Stranger6667:
The resulting workflow is similar to #564 , for example:
- 10 original Hypothesis tests for endpoint A
- 10 modified tests for endpoint A

And one important part - the modified tests will use completely different `case` instances than were used in the first 10 tests because these will be another hypothesis tests that will generate new cases from that `st.builds` strategy. I think that getting & applying that hook should happen inside `network_test` / `wsgi_test` so it is the same `case` instance. It will also not produce additional line in the CLI output (not sure if you wanted to have it separately in CLI or not)

src/schemathesis/runner/impl/solo.py
barrett-schonefeld: Yes I agree. I overlooked that several cases may be generated in `run_test`. I moved the hook to `network_test` and `wsgi_test`.

barrett-schonefeld:
@Stranger6667 

>  It will also not produce additional line in the CLI output (not sure if you wanted to have it separately in CLI or not)

I've been exploring what you're saying about the output. It seems that, once one of the cases experiences a failure, the other duplicated cases will not be validated. I will do my best to explain this below. Please let me know if I need to clear anything up in my explanation.

With the following check:

    @schemathesis.register_check
    def add_case_check(response, case):
        assert True

we get the following output:

    GET /api/path1 .                                                       [ 50%]
    GET /api/path2 .                                                       [100%]
    ...
    add_case_check                    2 / 2 passed          PASSED


However, with the following check:

    @schemathesis.register_check
    def add_case_check(response, case):
        assert False, "this check fails"

we get the following output. Notice 4 failures instead of 2 failures even using `--hypothesis-max-examples=1`:

    GET /api/path1 F                                                       [ 50%]
    GET /api/path2 F                                                       [100%]
    ...
    add_case_check                    0 / 4 passed          FAILED

When we scale this up to include `add_case` hooks, the success works as I expect. With the following hooks:

    @schemathesis.hooks.register("add_case")
    def add_first_header(context, case):
        if not case.headers:
            case.headers = {}
        case.headers["first"] = "first header"
        return case

    @schemathesis.hooks.register("add_case")
    def add_second_header(context, case):
        if not case.headers:
            case.headers = {}
        case.headers["second"] = "second header"
        return case

    @schemathesis.register_check
    def add_case_check(response, case):
        assert True

we get 6 successes, which is exactly what we expect because we have 3 cases x 2 endpoints x 1 check:

    GET /api/path1 .                                                        [ 50%]
    GET /api/path2 .                                                       [100%]
    ...
    add_case_check                    6 / 6 passed          PASSED

However, with the following hooks (`add_case` hooks stay the same, but note the `add_case_check` now has logic to fail with it sees the duplicated cases):

    @schemathesis.hooks.register("add_case")
    def add_first_header(context, case):
        if not case.headers:
            case.headers = {}
        case.headers["first"] = "first header"
        return case

    @schemathesis.hooks.register("add_case")
    def add_second_header(context, case):
        if not case.headers:
            case.headers = {}
        case.headers["second"] = "second header"
        return case

    @schemathesis.register_check
    def add_case_check(response, case):
        if (
            case.headers and 
            (
                case.headers.get("first") == "first header" or
                case.headers.get("second") == "second header"
            )
        ):
            assert False, "failing the added cases"

we get the following output. Notice in the `FAILURES` section, output is only provided for the Case object produced by the first `add_case` hook (the `Headers` indicate which Case it is):

    GET /api/path1 F                                                       [ 50%]
    GET /api/path2 F                                                       [100%]
    
    =================================== FAILURES ===================================
    ______________________________ GET: /api/path1 _______________________________
    1. failing the added cases
    
    Check           : add_case_check
    Headers         : {'first': 'first header'}
    
    Run this Python code to reproduce this failure: 
    
        requests.get('http://127.0.0.1:58094/api/failure', headers={'first': 'first header'})
    
    Or add this option to your command line parameters: --hypothesis-seed=290900904187938431298927631267103856427
    ______________________________ GET: /api/path2 _______________________________
    1. failing the added cases
    
    Check           : add_case_check
    Headers         : {'first': 'first header'}
    
    Run this Python code to reproduce this failure: 
    
        requests.get('http://127.0.0.1:58094/api/success', headers={'first': 'first header'})
    
    Or add this option to your command line parameters: --hypothesis-seed=294497505806281822851977016581346678738
    =================================== SUMMARY ====================================
    Performed checks:
        add_case_check                    4 / 8 passed          FAILED


The increased number of tests (when a failure occurs) seems to happen in the general case before `add_case` hooks were added.

The one downside is that we only get output from the Case that fails first, but I like the approach in this PR more than #564 because it allows more fine-grained control over the request and makes it easy to target specific API behavior. What do you think about the output and the behavior above? Do you think it makes it difficult to understand the results when we see (in the last example) that somehow 4 tests fail but only 2 failures are described in the `FAILURES` section?

src/schemathesis/runner/impl/core.py
```diff
@@ -197,15 +200,24 @@ def wsgi_test(
...
+            response = _case.call_wsgi(headers=headers)
+            elapsed = time.monotonic() - start
+        run_targets(targets, elapsed)
+        if store_interactions:
+            result.store_wsgi_response(_case, response, headers, elapsed)
+        result.logs.extend(recorded.records)
+        run_checks(_case, checks, result, response)
+
+
+def _get_cases(case: Case) -> Generator[Case, None, None]:
```
Stranger6667, line 215: I think we can do it in a slightly different way and avoid this extra function with `None` as a special value. If we'll factor out a function that should be called on `case` then it will be something like:

```
_network_test(case, ...)  # some args
context = HookContext(case.endpoint)
for case_hook in hooks.get_all_by_name("add_case"):
    _case = case_hook(context, deepcopy(case))
    _network_test(_case, ...)  # some args
```

And we won't need a condition and `None` inside a separate function. Also maybe the hook execution part can be extracted in a separate function as well and be parametrized by `_network_test` / `_wsgi_test` functions that will encapsulate logic from `network_test` / `wsgi_test`

src/schemathesis/models.py
```diff
@@ -189,6 +189,17 @@ def validate_response(
         if errors:
             raise AssertionError(*errors)
 
+    def __deepcopy__(self, memo: Optional[Dict[Any, Any]]) -> "Case":
+        return Case(
```
Stranger6667, line 193: Hm, is it intentional? As far as I understand, the first hook will modify the data, then the second hook will receive its modified version, which is probably not what we want - I assume that all these cases should be independent. Please, let me know if my assumption is correct

barrett-schonefeld: > I assume that all these cases should be independent.

Yes, that is correct. I added [this test](https://github.com/kiwicom/schemathesis/pull/590/files#diff-3a837d2bf3cc88920c7ba962d6d94138R834) to ensure that mutating the Case object in place does not affect other cases. 

Before I defined `__deepcopy__` in the Case class, `deepcopy(case)` resulted in the exception `Expecting property name enclosed in double quotes' in "TypeError: cannot serialize '_io.FileIO' object`. I think this exception occurs trying to `deepcopy` the `form_data` field.

barrett-schonefeld: @Stranger6667 this `deepcopy` problem is proving to be quite complicated. The sources of the `TypeError` are `Case.endpoint.app` and `Case.endpoint.schema.app`.

My current solution is to add a `__deepcopy__` function to the `Endpoint` class. This function does a `deepcopy` on most of the fields, shallow copies `Endpoint.app`, and uses `Endpoint.schema.clone()` (a shallow copy) to copy the `BaseSchema`. I will push this code for you to review.

The problem with this approach is that this is not a true deep copy yet this will be the generic way of deep copying the `Endpoint` class. If someone contributes code in the future that uses `deepcopy`, they may "successfully" and unknowingly get a partial deep copy of the `Endpoint`. The positive of this implementation is that it deep copies every field that we would reasonably expect the user to change.

The alternative approach is to implement a:
    
    class Case:
    ...
    def partial_deepcopy() -> "Case":
        endpoint_copy = Endpoint(
            headers=deepcopy(self.path_parameters),
            cookies=deepcopy(self.cookies),
            ...
            app=self.app
            schema=self.schema.clone()
        )
        return Case(
            endpoint=endpoint_copy
            path_parameters=deepcopy(self.path_parameters)
            ...
            app=self.app
        )

I think this `partial_deepcopy` function is probably better because it prevents any obscure logic errors using `deepcopy` on an Endpoint in the future.

Stranger6667: > The sources of the TypeError are Case.endpoint.app and Case.endpoint.schema.app.
Yep, generally, it is possible to re-construct such objects, but I don't think it worth the efforts. Basically it was one of the reasons I removed these things from "serialized" events that are yielded by the runner.

I think that the `partial_deepcopy` implementation is the way to go :) 

:+1:

src/schemathesis/runner/impl/core.py
```diff
@@ -234,12 +242,28 @@ def network_test(
     headers = headers or {}
     headers.setdefault("User-Agent", USER_AGENT)
     timeout = prepare_timeout(request_timeout)
-    response = case.call(session=session, headers=headers, timeout=timeout)
+    _network_test(case, checks, targets, result, session, timeout, store_interactions, headers, feedback)
+    add_cases(case, _network_test, checks, targets, result, session, timeout, store_interactions, headers, feedback)
+
+
+def _network_test(
+    _case: Case,
```
Stranger6667, line 250: Probably this variable can be named `case`

src/schemathesis/runner/impl/core.py
```diff
@@ -271,16 +295,30 @@ def wsgi_test(
 ) -> None:
     # pylint: disable=too-many-arguments
     headers = _prepare_wsgi_headers(headers, auth, auth_type)
+    _wsgi_test(case, checks, targets, result, headers, store_interactions, feedback)
+    add_cases(case, _wsgi_test, checks, targets, result, headers, store_interactions, feedback)
+
+
+def _wsgi_test(
+    _case: Case,
```
Stranger6667, line 303: And in this case, as well - there is no shadowing of another `case` variable, so we can keep it without the leading underscore

test/test_models.py
```diff
@@ -58,6 +59,36 @@ def test_call(override, base_url, swagger_20):
...
+        endpoint=endpoint,
+        path_parameters={"test": "test"},
+        headers={"Content-Type": "application/json"},
+        cookies={"TOKEN": "secret"},
+        query={"a": 1},
+        body={"b": 1},
+        form_data={"first": "John", "last": "Doe"},
+    )
+
+    copied_case = deepcopy(original_case)
```
Stranger6667, line 74: Hm, should it test `partial_deepcopy` instead?

test/test_models.py
```diff
@@ -58,6 +59,36 @@ def test_call(override, base_url, swagger_20):
     assert not records
 
 
+def test_case_deepcopy(swagger_20, openapi_30):
```
Stranger6667, line 62: `openapi_30` is not used. I assume that the idea is to test this behavior under both spec versions. If yes, then I propose to use a parametrized fixture:

```
@pytest.fixture(params=["swagger_20", "openapi_30"]
def schema(request):
    return request.getfixturevalue(request.param)
```

and then use `schema` fixture in the test

docs/customization.rst
```diff
@@ -188,3 +188,19 @@ With this simple handler only ``Done!`` will be displayed at the end of the test
...
+The user may change the Case object (and therefore the request data) before the request is sent to the server. The ``add_case`` allows the user to target specific
+behavior in the API by changing specific details of the duplicate request.
+
+.. code:: python
+
+    def add_case(context: HookContext, case: Case) -> Case:
+        case.headers["Content-Type"] = "application/json"
+        return case
+
+Note: A partial deep copy of the ``Case`` object is passed to each ``add_case`` hook. ``Case.endpoint.app`` and ``Case.endpoint.schema`` are shallow copied, so changes to
```
Stranger6667, line 205: I believe that with `Case.endpoint.app` it is adding a reference to the object rather than making a copy, but I am not completely sure. What do you think?

Stranger6667:
@barrett-schonefeld 

Sorry, I somehow missed this comment - https://github.com/kiwicom/schemathesis/pull/590#issuecomment-632193092

Let me know if my review comments clarify some points and if there is still something that I can provide feedback on :)

docs/customization.rst
barrett-schonefeld: That's a good point. In general, a shallow copy is the duplication of references to mutable objects. For `Case.endpoint.app`, it is the duplication of a reference to one mutable object, so it may be both. I updated the description to use "reference" for clarity.

test/test_models.py
barrett-schonefeld: I needed a schema to initialize the `Endpoint`. I initially planned to use both, but changed my plan and forgot to remove `openapi_30`.

barrett-schonefeld:
Thank you for working with me on this! I think I have tidied up the PR.

Stranger6667:
I'll work on a few more things and will make a new release :)

PR #606, opened 2020-05-29
feat: Support for parameters serialization formats in Open API 2 / 3

### LINKED ISSUE #599: Support different serialization styles for Open API parameters
As asked in Gitter we can support different parameter styles as defined [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#style-values).

Currently, we serialize some cases incorrectly. For example if we have a parameter defined like this:

```
{
    "description": "Parameters",
    "explode": False,
    "in": "query",
    "name": "abcd",
    "schema": {
        "additionalProperties": False,
        "properties": {"a": {"type": "number"}, "b": {"type": "number"}},
        "type": "object",
    },
    "style": "deepObject",
}
```

Then it is serialized like this - `/users?abcd=123`, however, it should be done differently

```
/users?abcd[a]=123&abcd[b]=456
```

Stranger6667:
Resolves #599 

TODO:
- [x] Support relevant formats for Open API 2
- [x] Other "in" locations - path, cookie, header
- [x] Docs & more tests
- [x] Serialize primitives

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,6 +12,7 @@ Added
 - Support for YAML files in references via HTTP & HTTP schemas. `#600`_
 - Stateful testing support via ``Open API links`` syntax. `#548`_
 - New ``add_case`` hook. `#458`_
+- Support for parameters serialization formats in Open API 2 / 3. For example ``pipeDelimited`` or ``deepObject``. `#599`_
 
 Changed
 ~~~~~~~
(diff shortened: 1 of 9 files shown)


src/schemathesis/specs/openapi/_hypothesis.py
```diff
@@ -0,0 +1,85 @@
...
+            item[name] = ",".join(item[name])
+        return item
+
+    return _map
+
+
+def make_comma_delimited_object(name: str) -> MapFunction:
+    def _map(item: Dict[str, Any]) -> Dict[str, Any]:
+        if name in item:
+            item[name] = ",".join(map(str, sum(item["color"].items(), ())))
```
Stranger6667, line 82: Fix hardcoded value

src/schemathesis/specs/openapi/_hypothesis.py
```diff
@@ -0,0 +1,85 @@
...
+                        conversions.append(make_space_delimited(name))
+                    if style is None or style == "form":  # "form" is the default style
+                        conversions.append(make_comma_delimited(name))
+
+    if conversions:
+        return compose(*conversions)
+    return None
+
+
+def make_deep_object(name: str) -> MapFunction:
```
Stranger6667, line 41: All these functions are similar and probably can be refactored

PR #611, opened 2020-06-08
feat: GraphQL support

Stranger6667:
WIP support for GraphQL. At the moment it is possible to create a strategy and use it in a test:

```python
schema = schemathesis.graphql.from_url("https://bahnql.herokuapp.com/graphql")

@given(case=schema.query.as_strategy()) 
@settings(deadline=None) 
def test(case): 
    response = case.call() 
    assert response.status_code < 500, response.content
```

Execution of the test above sometimes triggers `{"errors":[{"message":"Betriebsstellen: Failed to parse JSON","locations":[{"line":3,"column":5}],"path":["search","operationLocations"]}],"data":null}'` on a valid query :)

Roadmap:

- [ ] Design the API
- [ ] Integrate with pytest
- [ ] Integrate with CLI
- [ ] More complex test app to test nested objects, etc
- [ ] Mutations support
- [ ] Filters, hooks (?)
- [ ] Fix dependencies (there should be beta versions of packages)
- [ ] Custom scalar types support? Or at least handle gracefully - maybe exclude such a field from a query if possible

The current implementation on the `hypothesis-graphql` side is extremely slow :( I need to find ways to speed it up. 

Also, I am not sure that it is feasible to create a more generic interface in `BaseSchema` - maybe it will be easier to handle GraphQL things separately rather than trying to embed them into existing code - many things are not applicable here (e.g. tags, methods, endpoints, all `Case` structure, etc)

Comments, questions & suggestions are welcome

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ Schemathesis
 
 |Build| |Coverage| |Version| |Python versions| |Docs| |Chat| |License|
 
-Schemathesis is a tool for testing your web applications built with Open API / Swagger specifications.
+Schemathesis is a tool for testing your web applications built with Open API / Swagger or GraphQL specifications.
 
 It reads the application schema and generates test cases which will ensure that your application is compliant with its schema.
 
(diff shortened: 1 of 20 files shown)


Stranger6667:
I decided to add limited support as it is now and then split tasks and implement them separately

Issue #589, opened 2020-05-15
Support multiple examples specified in Open API schema


Stranger6667:
As mentioned in #582 we can generate multiple examples via the `examples` keyword:

```
parameters:
  - in: query
    name: limit
    schema:
      type: integer
      maximum: 50
    examples:       # Multiple examples
      zero:         # Distinct name
        value: 0    # Example value
        summary: A sample limit value  # Optional description
      max: # Distinct name
        value: 50   # Example value
        summary: A sample limit value
```

Stranger6667:
And I am very happy to hear that you'd like to implement it :) Let me know if I can provide support on this

barrett-schonefeld:
@Stranger6667 sounds great! I will start exploring the approaches you mentioned and will see your approach to GraphQL when you open a PR. Just to clarify, I plan to implement the version where we pair examples arbitrarily:

    request(limit=0, id=1), request(limit=50, id=2), request(limit=0, id=3)


Also, do we need to consider the possibility of property-level examples in request bodies as discussed in #467, or will #467 be implemented later?

Stranger6667:
I think that we can start with limited scope for now and expand it later :) But it will be good if we'll arrange the code for this feature in the way so it will be easier to extend it later with #467

barrett-schonefeld:
@Stranger6667 it looks like the `jsonschema` package does not support the `examples` keyword.

![Screen Shot 2020-06-05 at 8 13 35 AM](https://user-images.githubusercontent.com/59850348/83880799-56bb1e00-a705-11ea-880e-24085443b1fa.png)

I looked at the JSON Schema specification and found that it does have an `examples` keyword, but they treat the value of `examples` as an array. 


    "examples" : [
        "Anything",
        4035
    ]

I'm going to open an issue against `hypothesis-jsonschema`, and I hope they will consider supporting `examples without validating the value of `examples`.

Stranger6667:
Hm, it doesn't seem related to the `jsonschema` package, since we have JSON Schema definitions for Open API copied from the official sources, and there are `examples` keywords in the relevant places. What schema are you validating?

I am also not sure how `hypothesis-jsonschema` is related to it since it generates valid data for the passed schema. Could you, please, provide more info?

Stranger6667:
Based on [these definitions](https://github.com/kiwicom/schemathesis/blob/master/src/schemathesis/specs/openapi/definitions.py) we have the following

Open API 2:

[...]

PR #620, opened 2020-06-18
feat: support for multiple examples

barrett-schonefeld:
Implements #589

Updates:
- add specs/openapi/examples to hold the logic for retrieving examples from endpoints
- add support for OpenAPI examples

Tests:
- add unit tests for retrieving examples from "examples"
- add tests to make sure the static_parameter objects are created correctly

Note:
- Implementation allows the "example" and "examples" keywords to be used at the same time. We get all examples from "examples", and if "example" is used also, we create one more strategy using the "example" values.

### DIFF
--- a/src/schemathesis/_hypothesis.py
+++ b/src/schemathesis/_hypothesis.py
@@ -19,6 +19,14 @@ from .models import Case, Endpoint
 from .types import Hook
 
 PARAMETERS = frozenset(("path_parameters", "headers", "cookies", "query", "body", "form_data"))
+LOCATION_TO_CONTAINER = {
+    "path": "path_parameters",
+    "query": "query",
+    "header": "headers",
+    "cookie": "cookies",
+    "body": "body",
+    "formData": "form_data",
+}
 SLASH = "/"
 
 
@@ -57,31 +65,9 @@ def make_async_test(test: Callable) -> Callable:
     return async_run
 
 
-def get_example(endpoint: Endpoint) -> Optional[st.SearchStrategy[Case]]:
-    static_parameters = {}
-    for name in PARAMETERS:
-        parameter = getattr(endpoint, name)
-        if parameter is not None and "example" in parameter:
-            static_parameters[name] = parameter["example"]
-    if static_parameters:
-        strategies = {
-            parameter: prepare_strategy(
-                parameter, getattr(endpoint, parameter), endpoint.get_hypothesis_conversions(parameter)
-            )
-            for parameter in PARAMETERS - set(static_parameters)
-            if getattr(endpoint, parameter) is not None
-        }
-        return _get_case_strategy(endpoint, static_parameters, strategies)
-    return None
-
-
 def add_examples(test: Callable, endpoint: Endpoint, hook_dispatcher: Optional[HookDispatcher] = None) -> Callable:
     """Add examples to the Hypothesis test, if they are specified in the schema."""
-    strategy = get_example(endpoint)
-    if strategy is not None:
-        examples = get_single_example(strategy)
-    else:
-        examples = []
+    examples: List[Case] = [get_single_example(strategy) for strategy in endpoint.get_strategies_from_examples()]
     context = HookContext(endpoint)  # context should be passed here instead
     GLOBAL_HOOK_DISPATCHER.dispatch("before_add_examples", context, examples)
     endpoint.schema.hooks.dispatch("before_add_examples", context, examples)
@@ -92,7 +78,7 @@ def add_examples(test: Callable, endpoint: Endpoint, hook_dispatcher: Optional[H
     return test
 
 
-def get_single_example(strategy: st.SearchStrategy[Case]) -> List[Case]:
+def get_single_example(strategy: st.SearchStrategy[Case]) -> Case:
     @hypothesis.given(strategy)  # type: ignore
     @hypothesis.settings(  # type: ignore
         database=None,
@@ -107,7 +93,7 @@ def get_single_example(strategy: st.SearchStrategy[Case]) -> List[Case]:
 
     examples: List[Case] = []
     example_generating_inner_function()
-    return examples
+    return examples[0]
 
 
 def is_valid_header(headers: Dict[str, str]) -> bool:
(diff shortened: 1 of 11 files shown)


barrett-schonefeld:
One concern I have is that, when a request body has multiple `mediaTypes`, we retrieve `examples` from the "first" mediaType. However, because the API def is a dictionary, we don't know that calling `next(iter(media_types_dict))` actually gives the first `mediaType`.

The important thing is that we get `examples` from the same `mediaType` that is used in `specs/openapi/serialization`. But I don't know that we can rely on getting the same `mediaType` because different dictionaries may have different hashing functions.

src/schemathesis/specs/openapi/examples.py
```diff
@@ -0,0 +1,104 @@
...
+    return all_parameter_examples
+
+
+def get_request_body_examples(endpoint_def: Dict[str, Any], examples_field: str) -> Dict[str, Any]:
+    request_bodies_items = endpoint_def.get("requestBody", {}).get("content", {}).items()
+    if not request_bodies_items:
+        return {}
+    # first element in tuple in media type, second element is dict
+    media_type, schema = next(iter(request_bodies_items))
+    param_type = "body" if not media_type == "multipart/form-data" else "form_data"
```
Stranger6667, line 38: A more idiomatic way would be:
```suggestion
    param_type = "body" if media_type != "multipart/form-data" else "form_data"
```

src/schemathesis/specs/openapi/examples.py
```diff
@@ -0,0 +1,104 @@
...
+    static_parameters = {}
+    for name in PARAMETERS:
+        parameter = getattr(endpoint, name)
+        if parameter is not None and "example" in parameter:
+            static_parameters[name] = parameter["example"]
+    return static_parameters
+
+
+def get_static_params_from_examples(endpoint: Endpoint, examples_field: str) -> List[Dict[str, Any]]:
+    endpoint_def = endpoint.definition.raw
```
Stranger6667, line 52: Probably, in this case, we need to use `endpoint.definition.resolved` (that was added recently) because the definition for `examples` is `"oneOf": [{"$ref": "#/definitions/Reference"}, {"$ref": "#/definitions/Example"}]`

test/specs/openapi/test_examples.py
```diff
@@ -0,0 +1,139 @@
+import pytest
+import src.schemathesis as schemathesis
```
Stranger6667, line 2: Hm, I believe, that `import schemathesis` should work if it is tested via `tox` / a shell that has `PYTHONPATH` set to the `src` dir. It is done this way in other tests

src/schemathesis/specs/openapi/examples.py
```diff
@@ -0,0 +1,104 @@
...
+        if examples_field in param:
+            # add value to list of params with examples
+            param_type = {
+                "body": "body",
+                "cookie": "cookies",
+                "header": "headers",
+                "path": "path_parameters",
+                "query": "query",
+            }.get(param["in"])
+            assert param_type in PARAMETERS
```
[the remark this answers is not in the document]
barrett-schonefeld: My thought is that, if the keywords in `PARAMETERS` change over time but this mapping is not updated, the assertion would help catch the bug.

src/schemathesis/specs/openapi/examples.py
```diff
@@ -0,0 +1,104 @@
...
+    return None
+
+
+def merge_examples(
+    parameter_examples: List[Dict[str, Any]], request_body_examples: Dict[str, Any]
+) -> List[Dict[str, Any]]:
+    """Create list of static parameter objects from parameter and request body examples."""
+    static_param_list = []
+    for idx in range(num_examples(parameter_examples, request_body_examples)):
+        static_params: dict = {}
```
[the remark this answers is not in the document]
barrett-schonefeld: Yes!

src/schemathesis/specs/openapi/examples.py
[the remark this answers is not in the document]
Stranger6667: I think we can move this check to the import time - in this case it will be visible immediately if this mapping is up-to-date. Then this code might be rewritten as a list comprehension. Also, I am not sure but we might have it in multiple places (need to check this) so we can reuse if it will be available globally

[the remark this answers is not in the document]
barrett-schonefeld: Yes, the mapping is also used in `specs/openapi/links`. I will make the mapping global, removes the assertion above, and use the mapping in `links.py` and `examples.py`.

src/schemathesis/_hypothesis.py
```diff
@@ -19,6 +19,13 @@
 from .types import Hook
 
 PARAMETERS = frozenset(("path_parameters", "headers", "cookies", "query", "body", "form_data"))
+LOCATION_TO_CONTAINER = {
+    "path": "path_parameters",
+    "query": "query",
+    "header": "headers",
+    "cookie": "cookies",
+    "body": "body",
+}
```
Stranger6667, line 28: Do we need `form_data` here? E.g. if it might be triggered on Open API 2.0

test/specs/openapi/test_examples.py
```diff
@@ -0,0 +1,161 @@
...
+        "examples": [{"foo": "example1"}, {"foo": "example2"}, {"foo": "example3"}],
+    }
+    result = examples.merge_examples(parameter_examples, request_body_examples)
+
+    assert len(result) == 3
+    assert all("body" in static_params and "foo" in static_params["body"] for static_params in result)
+
+
+def test_examples_from_cli(app, testdir, cli, base_url, schema_with_examples):
+    uri = pathlib.Path(HERE).as_uri() + "/"
```
Stranger6667, line 151: This variable is not used later, seems like we can remove it completely

Stranger6667:
Great job! Re: dictionaries - I assume, that there might be corner-cases on non-CPython implementation of Python 3.6, but not sure how important it is to handle as we only declare CPython compatibility in `pyproject.toml`. Also, I don't know if PyPy has the same implementation detail as CPython for dicts on 3.6 (it might be, because that dicts impl came originally from PyPy) - otherwise, I am not sure if we need to support other implementations, that don't guarantee insertion ordering for dicts. I.e. I'd like to know what would be the affected space before taking action on this regard.

PR #625, opened 2020-06-19
feat: pass original case's response to add_case hook

barrett-schonefeld:
Updates:
- Allow user to not create an additional test case by returning None
- Pass the server response from the parent case to its child cases

Purpose:
- Allows user to create additional tests conditionally. For example, user may only want to create an additional test if the parent case received a 2xx response.

### DIFF
--- a/src/schemathesis/hooks.py
+++ b/src/schemathesis/hooks.py
@@ -9,6 +9,7 @@ from hypothesis import strategies as st
 
 from .models import Case, Endpoint
 from .types import GenericTest, Hook
+from .utils import GenericResponse
 
 
 def warn_deprecated_hook(hook: Hook) -> None:
@@ -231,10 +232,10 @@ def before_add_examples(context: HookContext, examples: List[Case]) -> None:
 
 
 @HookDispatcher.register_spec([HookScope.GLOBAL])
-def add_case(context: HookContext, case: Case) -> Case:
-    """Creates an additional test per endpoint.
+def add_case(context: HookContext, case: Case, response: GenericResponse) -> Optional[Case]:
+    """Creates an additional test per endpoint. If this hook returns None, no additional test created.
 
-    Called before request is sent with a copy of the original case object.
+    Called with a copy of the original case object and the server's response to the original case.
     """
 
 
--- a/src/schemathesis/runner/impl/core.py
+++ b/src/schemathesis/runner/impl/core.py
@@ -18,7 +18,7 @@ from ...runner import events
 from ...schemas import BaseSchema
 from ...stateful import ParsedData, StatefulTest
 from ...types import RawAuth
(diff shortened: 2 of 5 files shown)


src/schemathesis/hooks.py
```diff
@@ -231,10 +233,10 @@ def before_add_examples(context: HookContext, examples: List[Case]) -> None:
 
 
 @HookDispatcher.register_spec([HookScope.GLOBAL])
-def add_case(context: HookContext, case: Case) -> Case:
-    """Creates an additional test per endpoint.
+def add_case(context: HookContext, case: Case, case_response: Union[Response, WSGIResponse]) -> Union[None, Case]:
```
Stranger6667, line 236: We have `schemathesis.utils.GenericResponse` for `Union[Response, WSGIResponse]` and `Union[None, Case]` can be replaced with `Optional[Case]` which is more concise

src/schemathesis/runner/impl/core.py
```diff
@@ -235,11 +235,13 @@ def run_targets(targets: Iterable[Target], elapsed: float) -> None:
             hypothesis.target(elapsed, label="response_time")
 
 
-def add_cases(case: Case, test: Callable, *args: Any) -> None:
+def add_cases(case: Case, case_response: Union[requests.Response, WSGIResponse], test: Callable, *args: Any) -> None:
```
Stranger6667, line 238: `schemathesis.utils.GenericResponse` can be used here for `case_response`. Also, what do you think about using `response` name instead of `case_response`?

test/cli/test_commands.py
```diff
@@ -848,14 +876,14 @@ def test_multiple_add_case_hooks(testdir, cli, schema_url):
             import click
 
             @schemathesis.hooks.register("add_case")
-            def add_first_header(context, case):
+            def add_first_header(context, case, case_response):
```
Stranger6667, line 879: I think we can have `response` here as well

src/schemathesis/runner/impl/core.py
barrett-schonefeld: I like both ideas!

Stranger6667:
Looks great to me! only changelog entry is left

barrett-schonefeld:
@Stranger6667 my intuition says that users have not likely put many resources into creating `add_case` hooks yet. I would be OK merging the PR with the interface change, but I also understand if you want to be sure to deprecate the old interface and do the update properly.

Would we need to change the hook spec in `hooks.py`? And then in `core.py`:

      if len(inspect.signature(add_case).parameters) == 2:
          deprecated(...)(...)

Stranger6667:
> my intuition says that users have not likely put many resources into creating add_case hooks yet

I agree, probably it should be ok to have this change without backward compatibility

PR #629, opened 2020-06-21
Fix schema loader to parse numbers in scientific notation

tuffnatty:
`pyyaml` parses numbers in scientific notation without a dot as strings:
https://github.com/yaml/pyyaml/issues/173
https://github.com/yaml/pyyaml/pull/174

This PR presents a temporary solution until the issue is resolved upstream.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,11 @@ Changelog
 `Unreleased`_
 -------------
 
+Fixed
+~~~~~
+
+- Schema validation error on schemas containing numeric values in scientific notation without a dot. `#629`_
+
 `1.9.0`_ - 2020-06-20
 ---------------------
 
@@ -29,7 +34,6 @@ Fixed
 - Crash during generation of loosely-defined headers. `#621`_
 - Show exception information for test runs on invalid schemas with ``--validate-schema=false`` command-line option.
   Before the output sections for invalid endpoints were empty. `#622`_
-- Schema validation error on schemas containing numeric values in scientific notation without a dot. `#629`_
 
 `1.8.0`_ - 2020-06-15
 ---------------------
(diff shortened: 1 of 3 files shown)


test/test_loaders.py
```diff
@@ -62,3 +63,35 @@ def test_operation_id(operation_id):
     assert len(list(schema.get_all_endpoints())) == 1
     for endpoint in schema.get_all_endpoints():
         assert endpoint.definition.raw["operationId"] == operation_id
+
+
+@pytest.mark.hypothesis_nested
```
Stranger6667, line 68: This mark is not needed

Stranger6667: This mark can be removed

test/test_loaders.py
```diff
@@ -62,3 +63,35 @@ def test_operation_id(operation_id):
...
+                    ],
+                    "responses": {"200": {"description": "OK"}},
+                }
+            }
+        },
+    }
+
+    schema_path = testdir.makefile(".yaml", schema=json.dumps(schema))
+    # Then yaml loader should parse them without schema validation errors
+    schemathesis.from_path(str(schema_path))
```
Stranger6667, line 96: Could, you, please add some assertions about the parsed schema? e.g. that certain part it is not a string as was before

Stranger6667: For example, you can have something like this:

```
schema = schemathesis.from_path(str(schema_path))
value = schema.raw_schema["paths"][...][...]  # full path should be here
assert isinstance(value, float)
```

docs/changelog.rst
```diff
@@ -29,6 +29,7 @@ Fixed
 - Crash during generation of loosely-defined headers. `#621`_
 - Show exception information for test runs on invalid schemas with ``--validate-schema=false`` command-line option.
   Before the output sections for invalid endpoints were empty. `#622`_
+- Schema validation error on schemas containing numeric values in scientific notation without a dot. `#629`_
```
Stranger6667, line 32: Could you, please move it upper? This place is related to the `1.9.0` release - you can create a `Fixed` sub-section just after `Unreleased` section in the very top of this file

Stranger6667: It is around line 8

PR #512, opened 2020-04-17
WIP: feat: Support ASGI applications

HacKanCuBa:
This commit adds support to run ASGI applications directly from CLI
using `--app` option or in-code using `from_asgi`.

It is currently based on Starlette's TestClient.

WIP: App can be loaded correctly but can't run because new runners
needs to be created (along with support functions):

- SingleThreadASGIRunner
- ThreadPoolASGIRunner

PR #630, opened 2020-06-26
feat: Support ASGI applications

dongfangtianyu:
> This commit adds support to run ASGI applications directly from CLI
using --app option or in-code using from_asgi.
It is currently based on Starlette's TestClient.

Base on #512 
On the basis of this, we add some code and pass the preliminary test.

- `schemathesis.loaders.from_asgi` get schema
- `schemathesis.models.Case.call_asgi` using `starlette.testclient.TestClient` call API

### DIFF
--- a/src/schemathesis/models.py
+++ b/src/schemathesis/models.py
@@ -14,6 +14,7 @@ import attr
 import requests
 import werkzeug
 from hypothesis.strategies import SearchStrategy
+from starlette.testclient import TestClient as ASGIClient
 
 from .checks import ALL_CHECKS
 from .exceptions import InvalidSchema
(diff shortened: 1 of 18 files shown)


src/schemathesis/runner/__init__.py
```diff
@@ -276,6 +312,7 @@ def load_schema(
             # If `schema` is not an existing filesystem path or an URL then it is considered as an endpoint with
             # the given app
             loader = loaders.get_loader_for_app(app)
+            loader_options.update(dict_true_values(headers=headers))
```
Stranger6667, line 315: Hm, this change seems unrelated to the ASGI itself. On the first glance, it seems like we forgot to pass headers to loaders. Is it so? If yes, then it deserves a separate PR

dongfangtianyu: Yes, it is not must for asgi.

src/schemathesis/models.py
```diff
@@ -177,6 +178,23 @@ def call_wsgi(self, app: Any = None, headers: Optional[Dict[str, str]] = None, *
         with cookie_handler(client, self.cookies):
             return client.open(**data, **kwargs)
 
+    def call_asgi(
+        self,
+        app: Any = None,
+        base_url: Optional[str] = "http://testserver",
```
Stranger6667, line 184: Why is the default value `http://testserver` ? I'd rather keep it `None` for consistency with the `call` behavior unless it has a special meaning for ASGI apps

Stranger6667: Hm, it is indeed the default value for `starlette` - https://github.com/encode/starlette/blob/master/starlette/testclient.py#L369

src/schemathesis/runner/__init__.py
Stranger6667: Would you create a separate PR for that?

dongfangtianyu: Of course, it's a great honor. But the code test  failed on my other computer, and I'm looking for the reason, it may take some time

Stranger6667: Sure, thank you very much! :)

HacKanCuBa:
Thanks  @dongfangtianyu! :partying_face:

PR #687, opened 2020-07-28
fix: ensure the "examples" is non-empty before trying to index into it

barrett-schonefeld:


docs/changelog.rst
```diff
@@ -6,6 +6,12 @@ Changelog
 `Unreleased`_
 -------------
 
+Added
+~~~~~
+
+- Fix, check list is non-empty before indexing
```
Stranger6667, line 12: Could you please:

- rephrase it in a way so it reflects changes in user-facing behavior. Was it an error? If yes, then something like "`TypeError`, that happened during processing of empty "examples" list" (and put the proper error type there);
- Replace "Added" section name with "Fixed", since it is a bugfix;
- Remove one extra line after this entry

src/schemathesis/specs/openapi/examples.py
```diff
@@ -133,7 +133,7 @@ def merge_examples(
             static_parameters[parameter["type"]][parameter["name"]] = parameter["examples"][
                 min(idx, len(parameter["examples"]) - 1)
             ]
-        if "examples" in request_body_examples:
+        if "examples" in request_body_examples and len(request_body_examples["examples"]) > 0:
```
Stranger6667, line 136: What do you think about using `request_body_examples["examples"]` instead of  `len(request_body_examples["examples"]) > 0` which is more common for checking if a sequence is non-empty?

Stranger6667:
Thanks! I will issue a new release once the build is green

PR #691, opened 2020-08-03
feat: run checks conditionally

barrett-schonefeld:
@Stranger6667  We have custom checks that are only relevant in certain circumstances. For example, we have a check that only runs an assertion when the response status code is `406`. Currently, all checks run against all responses. This is OK and still correct, but the final results may be misleading. Consider the case where we send 100 requests during testing, and we only have one response with a status code of `406`. Lets assume the above-mentioned check (that only runs when the response status code is `406`) fails for the `406` response. Before, the output would say:

    406-check    99 / 100

but it would be more accurate to say:

    406-check    0 / 1

Note: I updated the return type of the checks to `Optional[bool]`. `MyPy`, however, failed unless the checks explicitly returned `None` even when the return type was marked `Optional`. Hence, I explicitly returned `None` for each of the checks.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -160,6 +160,21 @@ After registration, your checks will be available in Schemathesis CLI and you ca
 
     schemathesis --pre-run module.with.checks run -c new_check https://example.com/api/swagger.json
 
+Additionally, checks may return ``True`` to skip the check under certain conditions. For example, you may only want to run checks when the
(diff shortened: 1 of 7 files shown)


Stranger6667:
Sounds good to me! Could you, please add a changelog entry?

PR #724, opened 2020-09-18
feat: Stateful testing for the `pytest` runner

### LINKED ISSUE #616: Open API link testing support for Pytest
I really like the new feature for testing the Open API links. It seems like it is currently only supported via the command line interface and not with Pytest. Will this be added in the future or am I missing a config option here?

Stranger6667:
Working prototype

Resolves #616 

TODO:
- [x] Tests
- [x] Docs
- [x] Fix pytest percentage display
- [x] recursion limit support

### DIFF
--- a/src/schemathesis/_hypothesis.py
+++ b/src/schemathesis/_hypothesis.py
@@ -9,6 +9,7 @@ from .constants import DEFAULT_DEADLINE
 from .exceptions import InvalidSchema
 from .hooks import GLOBAL_HOOK_DISPATCHER, HookContext, HookDispatcher
 from .models import Case, Endpoint
+from .stateful import Feedback, Stateful
 
 
 def create_test(
@@ -16,7 +17,12 @@ def create_test(
 ) -> Callable:
     """Create a Hypothesis test."""
     hook_dispatcher = getattr(test, "_schemathesis_hooks", None)
-    strategy = endpoint.as_strategy(hooks=hook_dispatcher)
+    feedback: Optional[Feedback]
+    if endpoint.schema.stateful == Stateful.links:
+        feedback = Feedback(endpoint.schema.stateful, endpoint)
+    else:
+        feedback = None
+    strategy = endpoint.as_strategy(hooks=hook_dispatcher, feedback=feedback)
     wrapped_test = hypothesis.given(case=strategy)(test)
     if seed is not None:
         wrapped_test = hypothesis.seed(seed)(wrapped_test)
@@ -25,6 +31,7 @@ def create_test(
     setup_default_deadline(wrapped_test)
     if settings is not None:
         wrapped_test = settings(wrapped_test)
+    wrapped_test._schemathesis_feedback = feedback  # type: ignore
     return add_examples(wrapped_test, endpoint, hook_dispatcher=hook_dispatcher)
 
 
--- a/src/schemathesis/constants.py
+++ b/src/schemathesis/constants.py
@@ -9,3 +9,4 @@ except metadata.PackageNotFoundError:
 
 USER_AGENT = f"schemathesis/{__version__}"
 DEFAULT_DEADLINE = 500  # pragma: no mutate
+DEFAULT_STATEFUL_RECURSION_LIMIT = 5  # pragma: no mutate
--- a/src/schemathesis/extra/pytest_plugin.py
+++ b/src/schemathesis/extra/pytest_plugin.py
@@ -1,5 +1,5 @@
 from functools import partial
-from typing import Any, Callable, Generator, List, Optional, Type, cast
+from typing import Any, Callable, Generator, List, Optional, Type, Union, cast
 
 import pytest
 from _pytest import fixtures, nodes
@@ -7,12 +7,15 @@ from _pytest.config import hookimpl
 from _pytest.fixtures import FuncFixtureInfo
 from _pytest.nodes import Node
 from _pytest.python import Class, Function, FunctionDefinition, Metafunc, Module, PyCollector
+from _pytest.runner import runtestprotocol
+from _pytest.warning_types import PytestWarning
 from hypothesis.errors import InvalidArgument  # pylint: disable=ungrouped-imports
 from packaging import version
 
 from .._hypothesis import create_test
 from ..exceptions import InvalidSchema
 from ..models import Endpoint
+from ..stateful import Feedback
 from ..utils import is_schemathesis_test
 
 USE_FROM_PARENT = version.parse(pytest.__version__) >= version.parse("5.4.0")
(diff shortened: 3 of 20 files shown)


src/schemathesis/models.py
```diff
@@ -23,6 +23,7 @@
 
 if TYPE_CHECKING:
     from .hooks import HookDispatcher
+    from .runner.impl.core import Feedback
```
Stranger6667, line 26: Put `Feedback` to `schemathesis.stateful` to avoid cycles

src/schemathesis/models.py
```diff
@@ -39,6 +40,8 @@ class Case:
     body: Optional[Body] = attr.ib(default=None)  # pragma: no mutate
     form_data: Optional[FormData] = attr.ib(default=None)  # pragma: no mutate
 
+    feedback: "Feedback" = attr.ib(default=None)
```
Stranger6667, line 43: It shouldn't be visible in the repr, and it would be nice to have it as a private attribute

src/schemathesis/models.py
```diff
@@ -156,8 +159,12 @@ def call(
         response = session.request(**data)  # type: ignore
         if close_session:
             session.close()
+        self.store_response(response)
```
Stranger6667, line 162: Store wsgi as well (asgi is there via `call`)

src/schemathesis/extra/pytest_plugin.py
```diff
@@ -135,3 +168,13 @@ def pytest_pyfunc_call(pyfuncitem):  # type:ignore
...
     except InvalidArgument as exc:
         pytest.fail(exc.args[0])
+
+
+def pytest_runtest_protocol(item: Function, nextitem: Optional[Function]) -> bool:
+    item.ihook.pytest_runtest_logstart(nodeid=item.nodeid, location=item.location)
+    runtestprotocol(item, nextitem=nextitem)
+    item.ihook.pytest_runtest_logfinish(nodeid=item.nodeid, location=item.location)
+    if isinstance(item, SchemathesisFunction):
+        idx = item.session.items.index(item) + 1
```
Stranger6667, line 178: It could be isolated in `SchemathesisFunction`

src/schemathesis/extra/pytest_plugin.py
```diff
@@ -106,13 +117,35 @@ def collect(self) -> List[Function]:  # type: ignore
...
+                parent=self.parent,
+                callspec=getattr(self, "callspec", None),
+                callobj=test,
+                fixtureinfo=self._fixtureinfo,
+                keywords=self.keywords,
+                originalname=self.originalname,
+                test_func=self.test_function,
+            )
+            for (e, test) in self.obj._schemathesis_feedback.get_stateful_tests(
+                self.test_function, hypothesis.settings(), 1
```
Stranger6667, line 145: Pass proper settings & seed

src/schemathesis/extra/pytest_plugin.py
```diff
@@ -106,13 +117,35 @@ def collect(self) -> List[Function]:  # type: ignore
...
+                name=f"{self.originalname}[{e.method}:{e.full_path}]",
+                parent=self.parent,
+                callspec=getattr(self, "callspec", None),
+                callobj=test,
+                fixtureinfo=self._fixtureinfo,
+                keywords=self.keywords,
+                originalname=self.originalname,
+                test_func=self.test_function,
+            )
+            for (e, test) in self.obj._schemathesis_feedback.get_stateful_tests(
```
Stranger6667, line 144: `e` -> `endpoint`

src/schemathesis/extra/pytest_plugin.py
```diff
@@ -106,13 +117,35 @@ def collect(self) -> List[Function]:  # type: ignore
...
 
         This method is called only for this case.
         """
         return partial(self.obj, self.parent.obj)
 
+    def get_more_tests(self) -> List["SchemathesisFunction"]:
+        return [
+            create(
+                self.__class__,
+                name=f"{self.originalname}[{e.method}:{e.full_path}]",
```
Stranger6667, line 135: Improve the displayed name, we need to indicate that this test comes from the original one.

src/schemathesis/extra/pytest_plugin.py
```diff
@@ -53,7 +55,15 @@ def _gen_items(self, endpoint: Endpoint) -> Generator[Function, None, None]:
...
         if not metafunc._calls:
-            yield create(SchemathesisFunction, name=name, parent=self.parent, callobj=funcobj, fixtureinfo=fixtureinfo)
+            yield create(
+                SchemathesisFunction,
+                name=name,
+                parent=self.parent,
+                callobj=funcobj,
+                fixtureinfo=fixtureinfo,
+                test_func=self.test_function,
+                originalname=self.name,
```
Stranger6667, line 65: This line might be done separately - it doesn't affect the current state (to reduce the size of this PR)

src/schemathesis/constants.py
```diff
@@ -9,3 +11,7 @@
 
 USER_AGENT = f"schemathesis/{__version__}"
 DEFAULT_DEADLINE = 500  # pragma: no mutate
+
+
+class Stateful(enum.Enum):
```
Stranger6667, line 16: For now we can just use a string

src/schemathesis/_hypothesis.py
```diff
@@ -33,15 +36,22 @@ def create_test(
...
     if seed is not None:
         wrapped_test = hypothesis.seed(seed)(wrapped_test)
     if asyncio.iscoroutinefunction(test):
         wrapped_test.hypothesis.inner_test = make_async_test(test)  # type: ignore
     setup_default_deadline(wrapped_test)
     if settings is not None:
-        wrapped_test = settings(wrapped_test)
+        try:
+            wrapped_test = settings(wrapped_test)
+        except:  # TODO. fix re-wrapping
```
Stranger6667, line 52: We need to store the original test because another endpoint will have other examples, etc. After that, this try/except won't be needed

src/schemathesis/_hypothesis.py
```diff
@@ -33,15 +36,22 @@ def create_test(
 ) -> Callable:
     """Create a Hypothesis test."""
     hook_dispatcher = getattr(test, "_schemathesis_hooks", None)
-    strategy = endpoint.as_strategy(hooks=hook_dispatcher)
+    from .runner.impl.core import Feedback  # pylint: disable=import-outside-toplevel
```
Stranger6667, line 39: Move this import to the top of the file, once the `Feedback` class will be moved

src/schemathesis/_hypothesis.py
```diff
@@ -16,6 +16,9 @@
 from .models import Case, Endpoint
 from .specs.openapi._hypothesis import STRING_FORMATS
 
+if TYPE_CHECKING:
+    from .runner.impl.core import Feedback
```
Stranger6667, line 20: Remove this once the `Feedback` class will be moved

src/schemathesis/_hypothesis.py
```diff
@@ -33,15 +36,22 @@ def create_test(
 ) -> Callable:
     """Create a Hypothesis test."""
     hook_dispatcher = getattr(test, "_schemathesis_hooks", None)
-    strategy = endpoint.as_strategy(hooks=hook_dispatcher)
+    from .runner.impl.core import Feedback  # pylint: disable=import-outside-toplevel
+
+    feedback = Feedback("links", endpoint)
```
Stranger6667, line 41: The "stateful" parameter should be "links", created feedback only in this case. Or maybe use enum already?

Stranger6667: If so, then that enum should be used in the runner as well

src/schemathesis/models.py
```diff
@@ -156,8 +159,12 @@ def call(
...
         return response
 
+    def store_response(self, response: GenericResponse) -> None:
+        self.feedback.add_test_case(self, response)
```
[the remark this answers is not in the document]
Stranger6667: Also, it is not always available - e.g. when a strategy is used directly

Issue #688, opened 2020-07-29
Refactoring meta-issue


Stranger6667:
Over time it became harder to implement certain features and I'd like to focus on refactoring and various improvements to the codebase for some time.

The major problem I want to solve is GraphQL integration - currently, it is quite hacky, and implementing pytest & CLI support without hacks requires some major changes in the Schemathesis structure.

Here are the directions that could be taken in order to move faster with GraphQL:

- [ ] Replace `GenericResponse` with a common abstraction. There are multiple conditions scattered across the codebase that check if a variable with that type is one or another option. A common abstraction will isolate this code in one place;
- [ ] Make `Endpoint` model more generic. For GraphQL some parts should be adjusted - `as_strategy` should use `hypothesis_graphql` under the hood, `definition` attribute should not be bound to Open API, etc.
- [ ] `Case` is used everywhere, but for GraphQL there is a different class. Protocols could be used instead to avoid inheritance
- [ ] `endpoint` might be a misleading term - for GraphQL there is only "query", need to think about it.
- [ ]  `parametrize` accepts keywords, that don't make sense for GraphQL - it should be done differently in this case. They are also stored on a `BaseSchema` instance, which doesn't seem right

As the final result, I'd like to see CLI, runner and pytest integration that will not be tied to Open API specific structures, but will rather work with ones that will be generic and probably will delegate spec-specific logic to their internal components, i.e. will use composition approach.

As a side-note: Not everything should be tight to that `BaseSchema` subclasses - probably it could be just some "dispatch"-like functions, not sure yet

PR #729, opened 2020-09-22
refactor: Use structural subtyping instead of the nominal one

Stranger6667:
This is an initial draft for #688 

This PR serves mostly to simplify step-by-step work on this refactoring

### DIFF
--- a/src/schemathesis/protocols.py
+++ b/src/schemathesis/protocols.py
@@ -1,0 +1,148 @@
+"""Protocols that describe how concrete implementations should behave."""
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Sequence, Tuple, TypeVar
+
+import requests
+from hypothesis.strategies import SearchStrategy
+from typing_extensions import Protocol
+
+from .types import Body, Cookies, FormData, Headers, PathParameters, Query
+from .utils import GenericResponse, WSGIResponse
+
+if TYPE_CHECKING:
+    from .hooks import HookDispatcher
+    from .stateful import Feedback, Stateful, StatefulTest
+
+
+class CaseProtocol(Protocol):
+    endpoint: "EndpointProtocol"
+    path_parameters: Optional[PathParameters]
+    headers: Optional[Headers]
+    cookies: Optional[Cookies]
+    query: Optional[Query]
+    body: Optional[Body]
+    form_data: Optional[FormData]
+
+    @property
+    def path(self) -> str:
+        ...
+
+    @property
+    def full_path(self) -> str:
+        ...
+
+    @property
+    def method(self) -> str:
+        ...
+
+    @property
+    def base_url(self) -> Optional[str]:
+        ...
+
+    @property
+    def app(self) -> Any:
+        ...
+
+    @property
+    def formatted_path(self) -> str:
+        ...
+
+    def partial_deepcopy(self) -> "CaseProtocol":
+        ...
+
+    def as_requests_kwargs(self, base_url: Optional[str] = None) -> Dict[str, Any]:
+        ...
+
+    def as_text_lines(self) -> List[str]:
+        ...
+
+    def get_code_to_reproduce(self, headers: Optional[Dict[str, Any]] = None) -> str:
+        ...
+
+    def get_full_base_url(self) -> Optional[str]:
+        ...
+
+    def call(
+        self,
+        base_url: Optional[str] = None,
+        session: Optional[requests.Session] = None,
+        headers: Optional[Dict[str, Any]] = None,
+        **kwargs: Any,
+    ) -> requests.Response:
+        ...
+
+    def call_wsgi(self, app: Any = None, headers: Optional[Dict[str, str]] = None, **kwargs: Any) -> WSGIResponse:
+        ...
+
+    def call_asgi(
+        self,
+        app: Any = None,
+        base_url: Optional[str] = "http://testserver",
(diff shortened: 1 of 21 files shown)


src/schemathesis/protocols.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class EndpointProtocol(Protocol[C]):
+    path: str
+    method: str
+    definition: Any
+    schema: Any
+    app: Any
+    base_url: Optional[str]
+    path_parameters: Optional[PathParameters]
```
Stranger6667, line 94: These properties actually have a stricter type and probably misleading names - they are schemas for path_parameters, etc. For Open API they are `Dict[str, Any]` since should be JSON schema objects

src/schemathesis/protocols.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+C = TypeVar("C", bound=CaseProtocol)
+
+
+class EndpointProtocol(Protocol[C]):
+    path: str
+    method: str
+    definition: Any
+    schema: Any
```
Stranger6667, line 91: This should be a more narrow type

src/schemathesis/protocols.py
```diff
@@ -0,0 +1,141 @@
...
+        ...
+
+
+C = TypeVar("C", bound=CaseProtocol)
+
+
+class EndpointProtocol(Protocol[C]):
+    path: str
+    method: str
+    definition: Any
```
Stranger6667, line 90: This one also should be a more narrow type

src/schemathesis/protocols.py
```diff
@@ -0,0 +1,141 @@
...
+        ...
+
+    def get_full_url(self) -> str:
+        ...
+
+
+C = TypeVar("C", bound=CaseProtocol)
+
+
+class EndpointProtocol(Protocol[C]):
```
Stranger6667, line 87: Some docs & comments?

src/schemathesis/protocols.py
```diff
@@ -0,0 +1,141 @@
...
+from hypothesis.strategies import SearchStrategy
+from typing_extensions import Protocol
+
+from .hooks import HookDispatcher
+from .stateful import StatefulTest
+from .types import Body, Cookies, FormData, Headers, PathParameters, Query
+from .utils import GenericResponse, WSGIResponse
+
+
+class CaseProtocol(Protocol):
```
Stranger6667, line 14: This actually should be named "Case" and the currently named class "Case" should be "OpenAPICase". Similar to "EndpointProtocol"

src/schemathesis/protocols.py
```diff
@@ -0,0 +1,141 @@
...
+    endpoint: "EndpointProtocol"
+    path_parameters: Optional[PathParameters]
+    headers: Optional[Headers]
+    cookies: Optional[Cookies]
+    query: Optional[Query]
+    body: Optional[Body]
+    form_data: Optional[FormData]
```
Stranger6667, line 21: I think that attributes that contain actual data (e.g. `form_data`, `body`) should not be exposed in the protocol and actually used by the concrete implementation via some method (which is defined in the protocol)

src/schemathesis/protocols.py
```diff
@@ -0,0 +1,141 @@
...
+    definition: Any
+    schema: Any
+    app: Any
+    base_url: Optional[str]
+    path_parameters: Optional[PathParameters]
+    headers: Optional[Headers]
+    cookies: Optional[Cookies]
+    query: Optional[Query]
+    body: Optional[Body]
+    form_data: Optional[FormData]
```
Stranger6667, line 99: Similarly, these schemas could be hidden in concrete implementations

src/schemathesis/protocols.py
```diff
@@ -0,0 +1,141 @@
...
+    def get_hypothesis_conversions(self, location: str) -> Optional[Callable]:
+        ...
+
+    def get_stateful_tests(self, response: GenericResponse, stateful: Optional[str]) -> Sequence[StatefulTest]:
+        ...
+
+    def get_strategies_from_examples(self) -> List[SearchStrategy[C]]:
+        ...
+
+    def get_request_payload_content_types(self) -> List[str]:
```
Stranger6667, line 130: Should it be here? Seems like relevant only to Open API

src/schemathesis/protocols.py
```diff
@@ -0,0 +1,141 @@
...
+        ...
+
+    def prepare_multipart(self, form_data: Optional[FormData]) -> Tuple[Optional[List], Optional[Dict[str, Any]]]:
```
Stranger6667, line 133: Relevant only to Open API? if so, then it shouldn't be exposed

src/schemathesis/specs/openapi/checks.py
```diff
@@ -2,19 +2,17 @@
...
 from ...exceptions import get_response_type_error, get_schema_validation_error, get_status_code_error
+from ...protocols import CaseProtocol
 from ...utils import GenericResponse, are_content_types_equal, parse_content_type
 
-if TYPE_CHECKING:
-    from ...models import Case
 
-
-def status_code_conformance(response: GenericResponse, case: "Case") -> Optional[bool]:
+def status_code_conformance(response: GenericResponse, case: CaseProtocol) -> Optional[bool]:
```
Stranger6667, line 15: Should it be restricted to "OpenAPICase"? I.e. it should be covariant

PR #745, opened 2020-09-25
feat: Support for GraphQL schemas for the `pytest` runner

### LINKED ISSUE #649: GraphQL integration with pytest runner
Currently, we have only loaders, schema objects, and strategies. This is the next logical step

```
schema = schemathesis.graphql.from_url("...")

@schema.parametrize()
def test(case):
    ...

```

Stranger6667:
Resolves #649

TODO:
- [x] Docs
- [x] Test `get_code_to_reproduce`
- [x] Update `as_werkzeug_kwargs`

### DIFF
--- a/src/schemathesis/specs/graphql/schemas.py
+++ b/src/schemathesis/specs/graphql/schemas.py
@@ -1,44 +1,79 @@
 from functools import partial
+from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, cast
+from urllib.parse import urlsplit
 
 import attr
 import graphql
-import requests
 from hypothesis import strategies as st
+from hypothesis.strategies import SearchStrategy
 from hypothesis_graphql import strategies as gql_st
 
+from ...checks import not_a_server_error
+from ...hooks import HookDispatcher
+from ...models import Case, CheckFunction, Endpoint
 from ...schemas import BaseSchema
+from ...stateful import Feedback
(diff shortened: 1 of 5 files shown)


test/specs/graphql/test_basic.py
```diff
@@ -45,7 +45,7 @@ def test_raw_schema(graphql_schema):
 
 @pytest.mark.hypothesis_nested
 def test_query_strategy(graphql_schema):
-    strategy = graphql_schema.query.as_strategy()
+    strategy = graphql_schema["/graphql/"]["POST"].as_strategy()
```
Stranger6667, line 48: It should be "/graphql" - in the way it was loaded

src/schemathesis/specs/graphql/schemas.py
```diff
@@ -1,44 +1,61 @@
...
 
-    @property
-    def query(self) -> GraphQLQuery:
-        return GraphQLQuery(path=self.location or "", schema=self.schema)
+    def _get_base_path(self) -> str:
+        return cast(str, urlsplit(self.location).path)
+
+    def get_all_endpoints(self) -> Generator[Endpoint, None, None]:
+        yield Endpoint(
+            base_url=self.location, path=self.base_path, method="POST", schema=self, definition=None  # type: ignore
```
Stranger6667, line 47: Is there a way to detect the method used by the schema?

PR #756, opened 2020-10-01

hebertjulio:

### Description

### Checklist
- [x] All tests passing
- [x] Added a changelog entry


src/schemathesis/models.py
```diff
@@ -178,8 +179,12 @@ def call(
             close_session = False
 
         data = self.as_requests_kwargs(base_url)
+        if "headers" not in data or data["headers"] is None:
```
Stranger6667, line 182: `"headers" not in data` is redundant, since the `as_requests_kwargs` method always returns a dictionary with "headers" included

src/schemathesis/models.py
```diff
@@ -178,8 +179,12 @@ def call(
...
+            data.update({"headers": {}})
```
Stranger6667, line 183: Instead of using `update` with a single key, I'd prefer using `__setitem__` - `data["headers"] = {}`. It is more readable and slightly faster.

src/schemathesis/models.py
```diff
@@ -178,8 +179,12 @@ def call(
...
         if headers is not None:
-            data["headers"] = {**(data["headers"] or {}), **headers}
+            data["headers"].update(headers)
+        if "User-Agent" not in data["headers"]:
+            data["headers"].update({"User-Agent": USER_AGENT})
```
Stranger6667, line 187: Instead of using `update` with a single key, I'd prefer using `__setitem__` - `data["headers"]["User-Agent"] = USER_AGENT`. It is more readable and slightly faster

src/schemathesis/models.py
```diff
@@ -216,14 +221,25 @@ def as_werkzeug_kwargs(self, headers: Optional[Dict[str, str]] = None) -> Dict[s
...
+        **kwargs: Any
+    ) -> WSGIResponse:
         application = app or self.app
         if application is None:
             raise RuntimeError(
                 "WSGI application instance is required. "
                 "Please, set `app` argument in the schema constructor or pass it to `call_wsgi`"
             )
         data = self.as_werkzeug_kwargs(headers)
+        if "headers" not in data or data["headers"] is None:
```
Stranger6667, line 237: This branch is redundant - `as_werkzeug_kwargs` always returns a dictionary with "headers", which is a dictionary.

src/schemathesis/models.py
```diff
@@ -216,14 +221,25 @@ def as_werkzeug_kwargs(self, headers: Optional[Dict[str, str]] = None) -> Dict[s
...
+            data.update({"headers": {}})
+        if headers is not None:
```
Stranger6667, line 239: This branch is redundant as well - headers merging is already handled inside the `as_werkzeug_kwargs` method

src/schemathesis/models.py
```diff
@@ -216,14 +221,25 @@ def as_werkzeug_kwargs(self, headers: Optional[Dict[str, str]] = None) -> Dict[s
...
+            data["headers"].update(headers)
+        if "User-Agent" not in data["headers"]:
+            data["headers"].update({"User-Agent": USER_AGENT})
```
Stranger6667, line 242: Instead of using `update` with a single key, I'd prefer using `__setitem__` - `data["headers"]["User-Agent"] = USER_AGENT`. It is more readable and slightly faster

src/schemathesis/models.py
```diff
@@ -178,8 +179,12 @@ def call(
             close_session = False
 
         data = self.as_requests_kwargs(base_url)
+        if "headers" not in data or data["headers"] is None:
+            data.update({"headers": {}})
         if headers is not None:
-            data["headers"] = {**(data["headers"] or {}), **headers}
+            data["headers"].update(headers)
+        if "User-Agent" not in data["headers"]:
```
Stranger6667, line 186: Our "headers" value doesn't count with case sensitivity and if that dictionary contains "user-agent", then I am not sure which variant will be used by requests, but apparently it will be the last value:

```
In [2]: requests.get("https://httpbin.org/get", headers={"user-agent": "foo", "User-Agent": "bar"}).json()                                                                                                                                                                                                                                              
Out[2]: 
{'args': {},
 'headers': {'Accept': '*/*',
  'Accept-Encoding': 'gzip, deflate',
  'Host': 'httpbin.org',
  'User-Agent': 'bar',
  'X-Amzn-Trace-Id': 'Root=1-5f7585c3-5603d8fe64303f8e0441b5c7'},
 'origin': '0.0.0.0',
 'url': 'https://httpbin.org/get'}
```

We can leave it as it is for now and solve case-sensitivity in a separate issue since it also involves code in the "loaders" module and the "runner" package. In general I consider it as a bug and will create a bug report for it.

test/cli/output/test_default.py
```diff
@@ -12,6 +12,7 @@
 from schemathesis.cli.output.default import display_internal_error
 from schemathesis.runner.events import Finished, InternalError
 from schemathesis.runner.serialization import SerializedTestResult
+from schemathesis.constants import USER_AGENT
```
Stranger6667, line 15: This import is redundant and can be removed

test/cli/test_cassettes.py
```diff
@@ -153,7 +154,8 @@ async def test_replay(openapi_version, cli, schema_url, app, reset_app, cassette
 
 
 def test_multiple_cookies(base_url):
-    response = requests.get(f"{base_url}/success", cookies={"foo": "bar", "baz": "spam"})
+    headers = {"User-Agent": USER_AGENT}
```
Stranger6667, line 157: Could you please, add an explicit `assert` that the proper "User-Agent" header is present in `serialized["headers"]`?

test/extra/test_aiohttp.py
```diff
@@ -4,9 +4,12 @@
...
 from ..apps import _aiohttp
 
+from schemathesis.constants import USER_AGENT
+
 
 def test_exact_port():
     app = _aiohttp.create_openapi_app(("success", "failure"))
     run_server(app, 8999)
-    response = requests.get("http://localhost:8999/schema.yaml")
+    headers = {"User-Agent": USER_AGENT}
```
Stranger6667, line 13: Changes in this test don't seem relevant to the test purpose - it only checks if starting a server on some specific port works. Adding headers to this request is not relevant

test/runner/test_runner.py
```diff
@@ -595,7 +595,7 @@ def test_reproduce_code_with_overridden_headers(args, openapi3_base_url):
         expected = f"requests.get('http://localhost/api/failure', headers={headers})"
     else:
         expected = f"requests.get('{openapi3_base_url}/failure', headers={headers})"
-    assert after.result.checks[1].example.requests_code == expected
+    assert sorted(after.result.checks[1].example.requests_code) == sorted(expected)
```
Stranger6667, line 598: This value is a string - making a sorted list of characters and compare it is error-prone. I suggest comparing the exact strings, but if it still doesn't work - what is the error here?

test/test_async.py
```diff
@@ -82,7 +83,8 @@ async def users(request):
 async def test_(request, aiohttp_client, app, case):
     request.config.HYPOTHESIS_CASES += 1
     client = await aiohttp_client(app)
-    response = await client.request(case.method, f"/v1{case.formatted_path}", headers=case.headers, json=case.body)
+    headers = case.headers or {"User-Agent": USER_AGENT}
```
Stranger6667, line 86: Changes in this test don't seem relevant. Please, remove them or let me know if I missed something

test/test_models.py
```diff
@@ -149,7 +150,8 @@ def test_(case):
 
 @pytest.mark.endpoints()
 def test_response_from_requests(base_url):
-    response = requests.get(f"{base_url}/cookies")
+    headers = {"User-Agent": USER_AGENT}
```
Stranger6667, line 153: Could you please, add an explicit `assert` that the proper "User-Agent" header is present in `serialized.headers`?

docs/changelog.rst
```diff
@@ -6,6 +6,11 @@ Changelog
 `Unreleased`_
 -------------
 
+Fixed
+~~~~~
+
+- Set default User-Agent header in `Case.call`. `#717`_
```
Stranger6667, line 12: Could you please add ".. _#717: https://github.com/schemathesis/schemathesis/issues/717" string to the end of this file where other links are placed? It will make the rendered changelog point to the proper issue location.

Here is a suggestion - with the `rst` syntax, we need double quotes to make this block highlighted (with single quotes, it will be italic)

```suggestion
- Set default User-Agent header in ``Case.call``. `#717`_
```

Stranger6667 (changes requested):
Thank you for implementing this! I left some comment :) Also, could you please use "rebase" instead of "merge"? I'd like to avoid merge commits

Stranger6667:
Another small thing - to avoid that check failing, you need to install & run [pre-commit](https://pre-commit.com/), which will use [black](https://github.com/psf/black) to format the source code and "isort" to sort imports.

test/test_models.py
hebertjulio: requests dont return User-Agent in response.

Stranger6667: Indeed, then this change is not relevant for this test

src/schemathesis/models.py
Stranger6667: I'd rather limit the scope of this PR and solve #757 separately

Issue #706, opened 2020-09-09
Ability to specify a different port for the base_url


Stranger6667:
It might simplify the setup - currently, the user has to specify the whole base_url

Usecase:

```python
schema = schemathesis.from_uri(
    "http://myservice.example.com/swagger.json", 
    base_url="http://myservice.example.com:8000/swagger.json"
)
```

vs:

```python
schema = schemathesis.from_uri(
    "http://myservice.example.com/swagger.json", 
    port=8000
)
```

**Implementation notes**
- Add `port: Optional[int] = None` argument to the `from_uri` loader
- if there is no `base_url`, but `port` is specified, then take the original URL and add this port
- If there is `base_url` and `port`, then `base_url` should take precedence. I think we can even raise an exception in this case

On the other hand, I don't see much value in this feature, since it is quite a little shortcut, but brings some ambiguity, which might be resolved with extra validation.

Stranger6667:
As a note - @stannum-l works on this feature in #760

PR #760, opened 2020-10-01
fix(uri): allows specifying port for base_url

stannum-l:
This patch set adds a new port parameter to .from_uri() when a base_url
is not specified but a port is, it will take the original URL and adjust
the port to the specified one.

Closes: #706

### Description

### Checklist
- [ ] All tests passing
- [ ] Added a changelog entry

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,6 +12,7 @@ Changelog
 - New hook to pre-commit, rstcheck, as well as updates to documentation based on rstcheck `#734`_
 - New check for maximum response time and corresponding CLI option ``--max-response-time`` `#716`_
 - New field with information about executed checks in cassettes. `#702`_
+- New ``port`` parameter added to ``from_uri()`` method. `#706`_
 
 **Fixed**
 
@@ -1389,6 +1390,7 @@ Deprecated
 .. _#718: https://github.com/schemathesis/schemathesis/issues/718
 .. _#716: https://github.com/schemathesis/schemathesis/issues/716
 .. _#708: https://github.com/schemathesis/schemathesis/issues/708
+.. _#706: https://github.com/schemathesis/schemathesis/issues/706
 .. _#705: https://github.com/schemathesis/schemathesis/issues/705
 .. _#702: https://github.com/schemathesis/schemathesis/issues/702
 .. _#695: https://github.com/schemathesis/schemathesis/issues/695
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ importlib_metadata = { version = "^1.1", python = "<3.8" }
 werkzeug = ">0.16.0"
 junit-xml = "^1.9"
 starlette = "^0.13.2"
+yarl = "^1.6.0"
 
 [tool.poetry.dev-dependencies]
 coverage = "^5"
@@ -75,7 +76,7 @@ multi_line_output = 3
 default_section = "THIRDPARTY"
 include_trailing_comma = true
 known_first_party = "schemathesis"
-known_third_party = ["_pytest", "aiohttp", "attr", "click", "fastapi", "flask", "graphene", "graphql", "graphql_server", "hypothesis", "hypothesis_graphql", "hypothesis_jsonschema", "jsonschema", "junit_xml", "packaging", "pytest", "pytest_subtests", "requests", "schemathesis", "starlette", "urllib3", "werkzeug", "yaml"]
+known_third_party = ["_pytest", "aiohttp", "attr", "click", "fastapi", "flask", "graphene", "graphql", "graphql_server", "hypothesis", "hypothesis_graphql", "hypothesis_jsonschema", "jsonschema", "junit_xml", "packaging", "pytest", "pytest_subtests", "requests", "schemathesis", "starlette", "urllib3", "werkzeug", "yaml", "yarl"]
 
 [build-system]
 requires = ["poetry>=0.12"]
--- a/src/schemathesis/loaders.py
+++ b/src/schemathesis/loaders.py
@@ -10,6 +10,7 @@ from jsonschema import ValidationError
 from starlette.applications import Starlette
 from starlette.testclient import TestClient as ASGIClient
 from werkzeug.test import Client
+from yarl import URL
 
 from .constants import USER_AGENT
 from .exceptions import HTTPError
@@ -54,6 +55,7 @@ def from_uri(
     endpoint: Optional[Filter] = None,
     tag: Optional[Filter] = None,
     operation_id: Optional[Filter] = None,
+    port: Optional[int] = None,
     *,
     app: Any = None,
     validate_schema: bool = True,
@@ -63,6 +65,8 @@ def from_uri(
     headers = kwargs.setdefault("headers", {})
     if "user-agent" not in {header.lower() for header in headers}:
         kwargs["headers"]["User-Agent"] = USER_AGENT
+    if not base_url and port:
+        base_url = str(URL(uri).with_port(port))
     response = requests.get(uri, **kwargs)
     try:
         response.raise_for_status()
--- a/test/test_loaders.py
+++ b/test/test_loaders.py
@@ -1,6 +1,8 @@
 import json
 
 import pytest
+from requests.models import Response
+from yarl import URL
 
 import schemathesis
 from schemathesis.constants import USER_AGENT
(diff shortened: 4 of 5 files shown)


src/schemathesis/loaders.py
```diff
@@ -54,13 +55,18 @@ def from_uri(
...
     *,
     app: Any = None,
     validate_schema: bool = True,
     **kwargs: Any,
 ) -> BaseOpenAPISchema:
     """Load a remote resource and parse to schema instance."""
     kwargs.setdefault("headers", {}).setdefault("User-Agent", USER_AGENT)
+    if not base_url and port:
+        o = urlparse(uri)
+        loc = re.sub(r':(\d+)$', '', o.netloc)
```
Stranger6667, line 68: The `ParseResult` class provides `hostname` and `port` - I suggest using them instead of regex (it will require catching `ValueError` in some corner case).

For example, an URL like this one will be mangled - "http://2001:0db8:85a3:0000:0000:8a2e:0370:7334/openapi.json".
The base URL will end up with a different IPv6 address. E.g. with port 8081 it will be "2001:0db8:85a3:0000:0000:8a2e:0370:8081" instead of "[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:8081"

I also suggest adding handling & a test case for IPv6 addresses.

src/schemathesis/loaders.py
```diff
@@ -54,13 +55,18 @@ def from_uri(
...
+    port: Optional[int] = None,
     *,
     app: Any = None,
     validate_schema: bool = True,
     **kwargs: Any,
 ) -> BaseOpenAPISchema:
     """Load a remote resource and parse to schema instance."""
     kwargs.setdefault("headers", {}).setdefault("User-Agent", USER_AGENT)
+    if not base_url and port:
+        o = urlparse(uri)
```
Stranger6667, line 67: Let's use "parsed" as the variable name

test/test_loaders.py
```diff
@@ -38,6 +40,13 @@ def test_base_url_override(openapi3_schema_url, url):
     assert endpoint.base_url == "http://example.com"
 
 
+def test_port_override(openapi3_schema_url):
+    schema = schemathesis.from_uri(openapi3_schema_url, port=8080)
+    endpoint = next(schema.get_all_endpoints())
+    url = re.sub(r':(\d+)$', '', urlparse(openapi3_schema_url).netloc)
+    assert f"{url}:8080" in endpoint.base_url
```
Stranger6667, line 47: For simplicity, I suggest `assert endpoint.base_url == "http://127.0.0.1:8081/schema.yaml"` (later I'll add a separate function to avoid hardcoding the address)

Stranger6667 (changes requested):
Thank you for your contribution :) 

I left a couple of comments - please, take a look :)

Also, please add a changelog entry

src/schemathesis/loaders.py
stannum-l: Thanks - that's a good idea, didn't think of ipv6

Stranger6667:
As an idea - it might be easier to use [Yarl](https://yarl.readthedocs.io/en/latest/api.html#new-url-generation)

It seems like [there is already](https://yarl.readthedocs.io/en/latest/api.html#yarl.URL.with_port) a method we need:

```python
>>> URL('http://example.com:8888').with_port(9999)
URL('http://example.com:9999')
>>> URL('http://example.com:8888').with_port(None)
URL('http://example.com')
```

Stranger6667:
And regarding IPv6 addresses that are not enclosed - let's raise an error in this case since [RFC2732](http://www.ietf.org/rfc/rfc2732.txt) implies square brackets

src/schemathesis/loaders.py
stannum-l: This will be removed in favor of using `yarl`.

docs/changelog.rst
```diff
@@ -6,6 +6,8 @@ Changelog
 `Unreleased`_
 -------------
 
+- Add ``port`` parameter to ``from_uri()`` method. `#706`_
```
Stranger6667, line 9: Could you please add `.. _#706: https://github.com/schemathesis/schemathesis/issues/706` string to the end of this file where other links are placed? It will make the rendered changelog point to the proper issue location.

test/test_loaders.py
```diff
@@ -1,6 +1,7 @@
 import json
 
 import pytest
+from yarl import URL
```
Stranger6667, line 4: I believe we need to add `yarl` to direct dependencies via `poetry add yarl`

Stranger6667 (changes requested):
Looks awesome! Thank you! :) I left a couple of minor comments; please take a look. Also, you need to rebase - there are new entries in the changelog that conflict with this PR. You need to adjust it :)

Otherwise - we are good to go :)

test/test_loaders.py
stannum-l: Done

Stranger6667 (approved):
Awesome! Thank you @stannum-l for implementing this! :)

test/test_loaders.py
```diff
@@ -38,6 +39,22 @@ def test_base_url_override(openapi3_schema_url, url):
...
+    endpoint = next(schema.get_all_endpoints())
+    assert endpoint.base_url == "http://127.0.0.1:8081/schema.yaml"
+
+
+def test_port_override_with_ipv6(openapi3_schema_url):
+    url = URL(openapi3_schema_url)
+    parts = list(map(int, url.host.split(".")))
+    ipv6_host = "2002:{:02x}{:02x}:{:02x}{:02x}::".format(*parts)
+    ipv6 = url.with_host("[%s]" % ipv6_host)
+    schema = schemathesis.from_uri(ipv6, port=8081)
```
Stranger6667, line 53: Oh, I missed that the build is failing on this line (other failures come from a flaky test).

I think it could be enough to use mocking here instead of adapting the server. Something like

`mocked = mocker.patch("requests.request", wraps=simple_openapi)`

In the example above, the `simple_openapi` variable is the fixture defined in the root conftest file

Stranger6667, line 53: Oh, I missed that the build is failing on this line (other failures come from a flaky test).

I think it could be enough to use mocking here instead of adapting the server. Something like

`mocked = mocker.patch("requests.request", wraps=simple_openapi)`

In the example above, the `simple_openapi` variable is the fixture defined in the root conftest file

Stranger6667 (changes requested):

Stranger6667:
Also, another small note on the failing "Verify commit messages" check - I plan to squash all commits into a single one before merging and will put the missing category there :)

stannum-l:
Thanks for the quick feedback. Fixed the title of the PR and squashed the commits, but the unit test is still failing due to a SchemaValidation exception in the `get_endpoint()`. Will take a look again later, checked this in for now.

test/test_loaders.py
```diff
@@ -38,6 +40,27 @@ def test_base_url_override(openapi3_schema_url, url):
     assert endpoint.base_url == "http://example.com"
 
 
+def test_port_override(openapi3_schema_url):
+    schema = schemathesis.from_uri(openapi3_schema_url, port=8081)
+    endpoint = next(schema.get_all_endpoints())
+    assert endpoint.base_url == "http://127.0.0.1:8081/schema.yaml"
+
+
+def test_port_override_with_ipv6(openapi3_schema_url, simple_openapi, mocker):
```
stannum-l, line 49: not sure if `simple_openapi` is the payload to mock in `requests.get` - it appears L60 will fail with

```
>           raise InvalidSchema("Schema parsing failed. Please check your schema.") from exc
E           schemathesis.exceptions.InvalidSchema: Schema parsing failed. Please check your schema.
```

Any suggestion, or I can go through the schema code in greater details.

test/test_loaders.py
```diff
@@ -38,6 +40,27 @@ def test_base_url_override(openapi3_schema_url, url):
...
+    resp = Response()
+    resp.status_code = 200
+    resp._content = json.dumps(simple_openapi).encode("utf-8")
+    mocker.patch("requests.get", return_value=resp)
+
+    url = URL(openapi3_schema_url)
+    parts = list(map(int, url.host.split(".")))
+    ipv6_host = "2002:{:02x}{:02x}:{:02x}{:02x}::".format(*parts)
+    ipv6 = url.with_host("[%s]" % ipv6_host)
+    schema = schemathesis.from_uri(ipv6, validate_schema=False, port=8081)
```
Stranger6667, line 59: It fails due to passing URL instance instead of a string. It should be str(ipv6) instead

pyproject.toml
```diff
@@ -43,6 +43,8 @@ importlib_metadata = { version = "^1.1", python = "<3.8" }
 werkzeug = ">0.16.0"
 junit-xml = "^1.9"
 starlette = "^0.13.2"
+yarl = "^1.6.0"
+responses = "^0.12.0"
```
Stranger6667, line 47: I don’t see this dependency used :) seems like it could be removed

Stranger6667:
There was a PR merged recently and this one has conflicts - please rebase :)

pyproject.toml
stannum-l: Thanks - fixed in next patch

Stranger6667 (approved):

Stranger6667:
Amazing! Thank you very much @stannum-l ! :)

PR #769, opened 2020-10-02
feat(checks): Add custom check for max response time

### LINKED ISSUE #716: Add `--max-response-time` CLI option to verify the app response time
A custom check that will fail if the response time is greater than the specified one.

**Implementation notes**
- Create a new option for the CLI app. It should have `int` type and will mean the time in milliseconds
- Pass it down to the `runner.impl.core.run_checks`
- Pass `elapsed_time` to `run_checks` as well
- If `max_response_time` is not empty and the elapsed time is greater than `max_response_time`, then create an `AssertionError` exception, add it to `errors` and call `result.add_failure` with proper arguments (similar to already existing call in that function). If the time is less, then call `result.add_success`
- Docs
- Changelog entry

NOTE. In the future, it could be just a generic check as, e.g. `not_a_server_error`, but it will require some refactoring and will imply backward-incompatible changes

svtkachenko:

### Description
Added new custom check for request max response time and corresponding CLI option.
Closes #716 

### Checklist
- [x] All tests passing
- [x] Added a changelog entry


src/schemathesis/runner/impl/core.py
```diff
@@ -188,6 +196,13 @@ def run_checks(case: Case, checks: Iterable[CheckFunction], result: TestResult,
             errors.append(exc)
             result.add_failure(check_name, case, message)
 
+    if max_response_time:
+        if elapsed_time > max_response_time:
+            message = "Check max_response_time failed"
+            errors.append(AssertionError(message))
+            result.add_failure("max_response_time", case, message)
+        result.add_success("max_response_time", case)
```
Stranger6667, line 204: It should be in the `else` branch`

src/schemathesis/runner/impl/core.py
```diff
@@ -188,6 +196,13 @@ def run_checks(case: Case, checks: Iterable[CheckFunction], result: TestResult,
             errors.append(exc)
             result.add_failure(check_name, case, message)
 
+    if max_response_time:
+        if elapsed_time > max_response_time:
+            message = "Check max_response_time failed"
```
Stranger6667, line 201:
```suggestion
            message = "Check 'max_response_time' failed"
```

Stranger6667 (changes requested):

src/schemathesis/cli/__init__.py
```diff
@@ -92,6 +92,11 @@ def schemathesis(pre_run: Optional[str] = None) -> None:
 @click.option(
     "--checks", "-c", multiple=True, help="List of checks to run.", type=CHECKS_TYPE, default=DEFAULT_CHECKS_NAMES
 )
+@click.option(
+    "--max-response-time",
+    help="A custom check that will fail if the response time is greater than the specified one in milliseconds",
+    type=int,
```
Stranger6667, line 98: I'd add some range for this type - it should be a positive number

src/schemathesis/runner/impl/core.py
```diff
@@ -292,6 +322,7 @@ def _wsgi_test(
     headers: Dict[str, Any],
     store_interactions: bool,
     feedback: Feedback,
+    max_response_time: Optional[int] = None,
```
Stranger6667, line 325: I think that the default value could be removed - there is no reason for that

Stranger6667 (approved):

PR #771, opened 2020-10-02

hebertjulio:

### Description
Set default User-Agent header in Case.call.

### Checklist
- [x] All tests passing
- [x] Added a changelog entry


src/schemathesis/models.py
```diff
@@ -78,7 +79,7 @@ def formatted_path(self) -> str:
 
     def get_full_base_url(self) -> Optional[str]:
         """Create a full base url, adding "localhost" for WSGI apps."""
-        parts = urlsplit(self.base_url)
+        parts = urlsplit(self.base_url or "")
```
Stranger6667, line 82: It doesn't seem relevant to the topic of this PR. Why is it here?

src/schemathesis/models.py
```diff
@@ -176,10 +177,12 @@ def call(
             close_session = True
         else:
             close_session = False
-
         data = self.as_requests_kwargs(base_url)
+        data["headers"] = data["headers"] or {}
         if headers is not None:
-            data["headers"] = {**(data["headers"] or {}), **headers}
+            data["headers"].update(headers)
```
Stranger6667, line 183: There is a problem with updating this key - it references the original `headers` dictionary (if it is not None) that is in the `self.headers` attribute. So if you are updating it, then it mutates that dictionary - if it is used in some other places in the test after `case.call`, then it might not be desired to have it mutated. I suggest adding `data["headers"] = data["headers"].copy()` few lines above (after this line - `data["headers"] = data["headers"] or {}`)

In the future, I plan to refactor it and make `as_requests_kwargs` and `as_werkzeug_kwargs` return copies to avoid unintentional mutations

src/schemathesis/models.py
```diff
@@ -224,6 +227,11 @@ def call_wsgi(self, app: Any = None, headers: Optional[Dict[str, str]] = None, *
                 "Please, set `app` argument in the schema constructor or pass it to `call_wsgi`"
             )
         data = self.as_werkzeug_kwargs(headers)
+        data["headers"] = data["headers"] or {}
+        if headers is not None:
+            data["headers"].update(headers)
```
Stranger6667, line 232: These lines are not needed, since headers merging is handled in the `as_werkzeug_kwargs` method.

src/schemathesis/specs/graphql/schemas.py
```diff
@@ -59,7 +59,7 @@ def base_path(self) -> str:
         return self._get_base_path()
 
     def _get_base_path(self) -> str:
-        return cast(str, urlsplit(self.location).path)
+        return cast(str, urlsplit(self.location or "").path)
```
Stranger6667, line 62: This change doesn't seem relevant to the topic of this PR. Why is it changed here?

test/cli/test_cassettes.py
```diff
@@ -153,14 +154,16 @@ async def test_replay(openapi_version, cli, schema_url, app, reset_app, cassette
...
+    headers = {"User-Agent": USER_AGENT}
+    response = requests.get(f"{base_url}/success", cookies={"foo": "bar", "baz": "spam"}, headers=headers)
     request = Request.from_prepared_request(response.request)
     serialized = {
         "uri": request.uri,
         "method": request.method,
         "headers": request.headers,
         "body": {"encoding": "utf-8", "base64_string": request.body},
     }
+    assert USER_AGENT in serialized["headers"]["User-Agent"]
```
Stranger6667, line 166:
```suggestion
    assert serialized["headers"]["User-Agent"] == USER_AGENT
```

I think that comparing the whole string is better than comparing if a substring is present in this case

test/test_models.py
```diff
@@ -17,9 +18,10 @@ def test_path(swagger_20):
 @pytest.mark.parametrize("override", (False, True))
 @pytest.mark.parametrize("converter", (lambda x: x, lambda x: x + "/"))
 def test_as_requests_kwargs(override, server, base_url, swagger_20, converter):
+    headers = {"User-Agent": USER_AGENT}
     base_url = converter(base_url)
     endpoint = Endpoint("/success", "GET", {}, swagger_20)
-    kwargs = {"endpoint": endpoint, "cookies": {"TOKEN": "secret"}}
+    kwargs = {"endpoint": endpoint, "cookies": {"TOKEN": "secret"}, "headers": headers}
```
Stranger6667, line 24: I'd rather revert this change and verify that by default headers are updated with the "User-Agent" header

Stranger6667 (changes requested):

test/cli/test_cassettes.py
hebertjulio: It true, but serialized["headers"]["User-Agent"] is a tuple.

Stranger6667: You are right, my bad

src/schemathesis/models.py
hebertjulio: My PRs are failing in mypy due to an optional value being passed to this function

src/schemathesis/specs/graphql/schemas.py
hebertjulio: My PRs are failing in mypy due to an optional value being passed to this function

src/schemathesis/models.py
Stranger6667: Aha, I see. It might be caused by the recent changes in the pre-commit config - I will fix it soon

src/schemathesis/models.py
hebertjulio: Yes, it  true, I will remove these lines.

src/schemathesis/models.py
hebertjulio: I waiting you fix it?

src/schemathesis/models.py
hebertjulio: Ok

src/schemathesis/models.py
Stranger6667: Gimme 5 minutes and then you can rebase your branch

Stranger6667: Ok, now it should be fine - I reverted the change of the default Python version for pre-commit. Let me know if it works for you

test/cli/test_cassettes.py
hebertjulio: No problem :)

test/test_models.py
hebertjulio: as_requests_kwargs dont set default user agent

PR #774, opened 2020-10-02

hebertjulio:

### Description
Set default User-Agent header in Case.call.

### Checklist
- [x] All tests passing
- [x] Added a changelog entry

### DIFF
--- a/src/schemathesis/models.py
+++ b/src/schemathesis/models.py
@@ -16,6 +16,7 @@ import werkzeug
 from hypothesis.strategies import SearchStrategy
 from starlette.testclient import TestClient as ASGIClient
 
+from .constants import USER_AGENT
 from .exceptions import InvalidSchema
 from .types import Body, Cookies, FormData, Headers, PathParameters, Query
 from .utils import GenericResponse, WSGIResponse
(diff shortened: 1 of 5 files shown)


src/schemathesis/models.py
```diff
@@ -176,10 +177,13 @@ def call(
...
         else:
             close_session = False
-
         data = self.as_requests_kwargs(base_url)
+        data["headers"] = data["headers"] or {}
+        data["headers"] = data["headers"].copy()
         if headers is not None:
-            data["headers"] = {**(data["headers"] or {}), **headers}
+            data["headers"].update(headers)
+        if "user-agent" not in set(k.lower() for k in data["headers"]):
```
Stranger6667, line 185: I'd prefer to keep the current behavior for consistency and limiting the scope of this PR

```suggestion
        if "User-Agent" not in data["headers"]:
```

src/schemathesis/models.py
```diff
@@ -224,6 +228,8 @@ def call_wsgi(self, app: Any = None, headers: Optional[Dict[str, str]] = None, *
                 "Please, set `app` argument in the schema constructor or pass it to `call_wsgi`"
             )
         data = self.as_werkzeug_kwargs(headers)
+        if "user-agent" not in set(k.lower() for k in data["headers"]):
```
Stranger6667, line 231:
```suggestion
        if "User-Agent" not in data["headers"]:
```

Stranger6667 (changes requested):

Stranger6667:
However, I think that I'll amend the commit a bit for simplicity.

All good! :) Merging

Issue #737, opened 2020-09-24
Alternative API for stateful tests in Python


Stranger6667:
Even though now all stateful tests are separate pytest items it might be better to run all of them in a single test function. Something like this:

```python
@schema.parametrize()
def test_api(case):
    for step in case.get_stateful_tests():
        response = step.call()
        step.validate_response(response)
```

It might be much easier to implement since there will be no need to integrate it with pytest itself. Also - there will be no mangled execution statistic in the pytest output because it is a single test, there will be no problems with changing hypothesis settings in subsequent steps, etc

Zac-HD:
Have you seen [REST-ler](https://www.microsoft.com/en-us/research/uploads/prod/2018/04/restler.pdf)?  It doesn't look like it's open source, but there are some nice ideas about inferring dependencies and order of calls that could enhance stateful mode :slightly_smiling_face:

Stranger6667:
Yep, I actually created this issue after reading that paper :) The current implementation with pytest is quite clunky - it modifies some internals between test executions. I hope that this implementation will be re-iterated in the next major release to incorporate ideas from REST-let & #520.

> Building schemathesis' stateful mode on top of Hypothesis' RuleBasedStateMachine would be really nice

I tried it a couple of times but decided to go with a more straightforward approach - incorporating RuleBasedStateMachine will require some significant refactorings, but I think it worth it in the long run :)

> Longer-term that would even allow coverage-guided fuzzing for webapps...

That sounds really cool! I ran some experiments with Hypothesis' `target,` and it was promising :) But it is still a huge area to research

Stranger6667:
As a sidenote

HypoFuzz looks awesome to me :) Wondering what bugs were found in `hypothesis-jsonschema` as mentioned in the text above the dashboard :)

Zac-HD:
- https://github.com/Julian/jsonschema/issues/743
- `{'not': {'anyOf': [{'type': 'integer'}, {'type': 'number'}]}}` should canonicalise to `{'type': ['null', 'boolean', 'string', 'array', 'object']}`, but produces an intermediate schema (which does) instead.  Similar to the known issue https://github.com/Zac-HD/hypothesis-jsonschema/issues/64, if less serious.
- Something about numeric bounds that I've had trouble reproducing; could be a bug in my test harness.

Stranger6667:
Implemented in #805 :tada:

Issue #584, opened 2020-05-12
Single JSON schema for all endpoint properties?


Stranger6667:
When I was experimenting with Open API links I found out that to work in a more reliable way it may require a single JSON schema for all endpoint parts - `headers`, `query`, `body`, etc.

For example, we want to run additional tests on endpoint X and take some data from the responses from endpoint Y. To enable shrinking we need to generate these new requests, thus simple replacing values in e.g. `query` will not work. Then we reuse our implementation where `st.builds` generates instances of `Case` from multiple components. These components are independent strategies themselves which leads to a problem:
1. We want to take data from different part of a response / request
2. We want all of them to be present in a generated data
3. It can't be done since `st.builds` arguments are independent

In other words if we have:
- requests query params: `[{"q": 1}, {"q": 2}, {"q": 3}]`
- responses bodies: `[{"user_id": 5}, {"user_id": 6}, {"user_id": 7}]
Then we want that generated requests contain certain combinations that were present in the same request/response:
- Query / body pair is one of `[{"q": 1}, {"user_id": 5}]`, `[{"q": 2}, {"user_id": 6}]`, `[{"q": 3}, {"user_id": 7}]`

The solution may be creating a single JSON schema and generate components as properties of this schema and then build `Case` from them. In this case, we can use `oneOf` to define needed combos

Stranger6667:
Not relevant, since #805 solves it

PR #805, opened 2020-10-12
feat: Stateful testing via Hypothesis's RuleBasedStateMachine

Stranger6667:
A draft for #737. It seems like this approach is much more flexible:
- Arbitrary action sequences. E.g. it could be calling some endpoints more than once - broader test surface.
- Adding the "source" attribute is trivial, since it is done in a "flatmap" strategy - just attach an attribute. With the current approach, it is much harder since it involves generating a new endpoint that contains a bunch of schemas in "anyOf" - there is no connection with the original response. So it is not clear what parameter came from what response - it might be done with attaching some extra mapping and passing it along until the case creation, which is quite hacky.
- Different parameter types can be carried together. Currently, all parameter types are chosen independently - changing is problematic (#584 )
- It doesn't have a problem when the deeper the call goes, the fewer parameter options are possible. It is just a different approach without this problem. The current approach behaves like this: 1st "POST" call returns 50 successful responses that are suitable for Open API link, then there are 50 options for e.g. subsequent "GET" call, then these 50 calls might return fewer successful responses for the next level, etc, so the number is decreasing. With the new approach, new POST calls may be done at any time and the number of "source" responses grows.
- It is much easier to implement multiple sources for a call - it boils down to using multiple bundles as a source for the `flatmap` call.

CLI integration probably in a separate PR.

### DIFF
--- a/src/schemathesis/specs/openapi/stateful/links.py
+++ b/src/schemathesis/specs/openapi/stateful/links.py
@@ -1,0 +1,80 @@
+from typing import TYPE_CHECKING, Callable, Dict, List, Tuple
+
+import hypothesis.strategies as st
+from requests.structures import CaseInsensitiveDict
+
+from ....stateful import StepResult
+from ..links import OpenAPILink, get_all_links
+from ..utils import expand_status_code
+
+if TYPE_CHECKING:
+    from ....models import Endpoint
+
+FilterFunction = Callable[[StepResult], bool]
+
+
+def apply(
+    endpoint: "Endpoint",
+    bundles: Dict[str, CaseInsensitiveDict],
+    connections: Dict[str, List[st.SearchStrategy[Tuple[StepResult, OpenAPILink]]]],
+) -> None:
+    """Gather all connections based on Open API links definitions."""
+    all_status_codes = list(endpoint.definition.resolved["responses"])
+    for status_code, link in get_all_links(endpoint):
+        target_endpoint = link.get_target_endpoint()
+        strategy = bundles[endpoint.path][endpoint.method.upper()].filter(
+            make_response_filter(status_code, all_status_codes)
+        )
+        connections[target_endpoint.verbose_name].append(_convert_strategy(strategy, link))
+
+
+def _convert_strategy(
+    strategy: st.SearchStrategy[StepResult], link: OpenAPILink
+) -> st.SearchStrategy[Tuple[StepResult, OpenAPILink]]:
+    # This function is required to capture values properly (it won't work properly when lambda is defined in a loop)
+    return strategy.map(lambda out: (out, link))
+
+
+def make_response_filter(status_code: str, all_status_codes: List[str]) -> FilterFunction:
+    """Create a filter for stored responses.
+
+    This filter will decide whether some response is suitable to use as a source for requesting some endpoint.
+    """
+    if status_code == "default":
+        return default_status_code(all_status_codes)
+    return match_status_code(status_code)
+
+
+def match_status_code(status_code: str) -> FilterFunction:
+    """Create a filter function that matches all responses with the given status code.
+
+    Note that the status code can contain "X", which means any digit.
+    For example, 50X will match all status codes from 500 to 509.
+    """
+    status_codes = set(expand_status_code(status_code))
+
+    def compare(result: StepResult) -> bool:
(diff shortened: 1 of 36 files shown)


src/schemathesis/specs/openapi/stateful.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def create_strategy(
+    bundle: Bundle, status_code: str, status_codes: List[str], target_strategy: SearchStrategy["Case"]
+) -> SearchStrategy["Case"]:
+    """Create a Hypothesis strategy for an endpoint.
+
+    The cases are generated based on responses that are considered "sources" for this endpoint.
+    """
+    if status_code == "default":
```
Stranger6667, line 82: It would be better if this part will be unaware of that specific filtering logic - it could be passed as a dependency. This way we can decide how to take responses upfront - e.g. links, inference, randomly, etc.

src/schemathesis/specs/openapi/stateful.py
```diff
@@ -0,0 +1,128 @@
...
+    yield f"rule {name}", rule(target=bundles[name], case=strategy)(call_and_validate)
+
+
+def create_strategy(bundles: Dict[str, Bundle], sources: Sources) -> SearchStrategy["Case"]:
+    """Create a Hypothesis strategy for an endpoint.
+
+    The cases are generated based on responses that are considered "sources" for this endpoint.
+    """
+    strategies = [
+        bundles[source_name].filter(response_filter).flatmap(strategy_builder)
```
Stranger6667, line 89: It could be a function, that takes all bundles and creates a new strategy - this way it could be more flexible - taking one bundle, multiple bundles to combine two responses, etc.

src/schemathesis/specs/openapi/stateful/links.py
```diff
@@ -0,0 +1,128 @@
...
+        old_case, response = item
+        context = expressions.ExpressionContext(case=old_case, response=response)
+
+        @st.composite  # type: ignore
+        def new_case(draw: st.DataObject) -> "Case":
+            case = draw(target_strategy)
+
+            for location, name, expression in link_parameters:
+                value = expressions.evaluate(expression, context)
+                if location:
```
Stranger6667, line 73: Move logic to get the container into a separate function

src/schemathesis/specs/openapi/stateful/links.py
```diff
@@ -0,0 +1,128 @@
...
+
+        old_case, response = item
+        context = expressions.ExpressionContext(case=old_case, response=response)
+
+        @st.composite  # type: ignore
+        def new_case(draw: st.DataObject) -> "Case":
+            case = draw(target_strategy)
+
+            for location, name, expression in link_parameters:
+                value = expressions.evaluate(expression, context)
```
Stranger6667, line 72: It could be done after getting the container

src/schemathesis/specs/openapi/stateful/links.py
```diff
@@ -0,0 +1,128 @@
...
+        except ValueError:
+            return None, parameter, expression
+
+    link_parameters = [
+        expand_parameter(parameter, expression) for parameter, expression in link.get("parameters", {}).items()
+    ]
+    link_request_body = link.get("requestBody", NOT_SET)
+
+    # Create a builder for strategies that will generate cases for the given endpoint
+    def make_strategy(item: Tuple["Case", requests.Response]) -> st.SearchStrategy["Case"]:
```
Stranger6667, line 62: Maybe it is better to create a special class, that holds all the needed context? and return its instance from `call_and_validate`?, then it will be easier to access its components

src/schemathesis/specs/openapi/stateful/links.py
```diff
@@ -0,0 +1,128 @@
...
+
+def make_links_strategy(
+    bundle: Bundle, target_endpoint: "Endpoint", link: Dict[str, Any], status_code: str, all_status_codes: List[str]
+) -> st.SearchStrategy["Case"]:
+    """Create a single Open API Links based strategy.
+
+    It combines all links that point to an endpoint into a single strategy.
+    Only one link is used at once - responses are not combined with each other.
+    """
+    if status_code == "default":
```
Stranger6667, line 41: Separate function?

PR #806, opened 2020-10-13
feat: generate cURL from Case instance

### LINKED ISSUE #689: Generate cURL command for a `Case` instance
As mentioned in #472

```python

@schema.parametrize()
def test_api(case):
    command = case.as_curl_command()
    ...
```

It should generate a string that could be copy-pasted and run in a terminal. The implementation itself might be similar to `Case.get_code_to_reproduce`, and the result should be compatible with the latest cURL behavior.

https://github.com/IBM/service-validator/pull/12 could be used as inspiration

ZhukovGreen:
Closes #689 

### Description
Add a method `as_curl_command` to the Case class to generate a string with the curl command for manual testing 

### Checklist
- [x] All tests passing
- [x] Added a changelog entry
- [x] Extended the README / documentation, if necessary

### DIFF
--- a/src/schemathesis/models.py
+++ b/src/schemathesis/models.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, Generator, Iterator, List
 from urllib.parse import urljoin, urlsplit, urlunsplit
 
 import attr
+import curlify
 import requests
 import werkzeug
 from hypothesis.strategies import SearchStrategy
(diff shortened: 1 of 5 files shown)


test/test_case_curl_cmd.py
```diff
@@ -0,0 +1,12 @@
...
+import schemathesis
+from schemathesis import Case
+
+schema = schemathesis.from_dict(app.openapi())
+
+
+@schema.parametrize()
+def test_basic_curl_gen(case: Case):
+    res = case.as_curl_command()
+    assert res == "curl -X GET " "-H 'User-Agent: schemathesis/2.6.0' " "http://localhost/users"
```
Stranger6667, line 12:
```suggestion
    command = case.as_curl_command()
    assert command == "curl -X GET -H 'User-Agent: schemathesis/2.6.0' http://localhost/users"
```

Stranger6667: There is also `schemathesis.constants.USER_AGENT` constant that could be used to format the expected string to avoid this test breaking after a version bump

src/schemathesis/models.py
```diff
@@ -126,6 +127,17 @@ def get_code_to_reproduce(self, headers: Optional[Dict[str, Any]] = None) -> str
             args_repr += f", {printed_kwargs}"
         return f"requests.{method}({args_repr})"
 
+    def as_curl_command(self, headers: Optional[Dict[str, Any]] = None) -> str:
+        """Construct a curl cmd for a given case."""
```
Stranger6667, line 131:
```suggestion
        """Construct a curl command for a given case."""
```

Stranger6667 (changes requested):
Awesome! Thank you for your contribution :) I left a couple of minor comments :)

Please, address them, add an entry to the `changelog.rst` and we are ready to merge :) Then I will add a new section to the docs about reproducing failures, that will include this feature as well :)

Stranger6667 (approved):

PR #831, opened 2020-10-25
feat: Negative testing

### LINKED ISSUE #65: Negative testing
The current approach produces data that **matches** the schema. To verify that wrong data will not be processed without and error we need to mutate the schema in runtime before creating a strategy in the way that it will not match the initial schema (e.g. it shouldn't be a subset of the initial schema). Could be another "test_type" option.

With this feature, we will expand the scope of errors that could be discovered by Schemathesis

### LINKED ISSUE #1152: [BUG] Avoid internal errors on unusual schemas for `multipart/form-data` media types
**Describe the bug**
If the API schema has the `multipart/form-data` but its schema is not of type "object", then Schemathesis fails with `AttributeError: 'str' object has no attribute 'items'`.

Multipart requests imply some structure in the payload, but some schemas do not define any structure (e.g. `{"type": "string"}`). 
Usually, they have the `object` type (e.g., in [docs](https://swagger.io/docs/specification/describing-request-body/multipart-requests/)), but I can't find any restriction on that matter in the spec. In any case, Schemathesis should not fail with an internal error.

**To Reproduce**
Run Schemathesis against the schema below

A modified version of the `multipart` endpoint from the test suite:

```yaml
components:
  securitySchemes:
    api_key:
      in: header
      name: X-Token
      type: apiKey
    basicAuth:
      scheme: basic
      type: http
info:
  description: An API to test Schemathesis
  title: Example API
  version: 1.0.0
openapi: 3.0.2
paths:
  /multipart:
    post:
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: string
        required: true
      responses:
        '200':
          description: OK
servers:
- url: https://127.0.0.1:8888/{basePath}
  variables:
    basePath:
      default: api
```

**Expected behavior**
I think that Schemathesis can still send multipart requests even with such schemas. For example, if some random string is generated, then we can wrap it in boundaries manually like this:

```
--12345
random-string
--12345
```

and supply a header like this - `{"Content-Type": "multipart/form-data; boundary=12345"}`

To make it work, we need to change a few things:

- If the [multipart serializer](https://github.com/schemathesis/schemathesis/blob/1050b81abdd6a51f6c67d35d5ec83758cda8a678/src/schemathesis/serializers.py#L157) receives not `bytes` or `dict`, then generate a random boundary value (via e.g. `urllib3.filepost.choose_boundary`) and build content like in the example above;
- Craft the proper `Content-Type` header inside this serializer and return it inside a dictionary. The return value looks somewhat like this - `{"data": multipart_content, "headers": {"Content-Type": multipart_header}}`;
- Inside `models.Case.as_requests_kwargs` assume that the `extra` value may contain the "headers" key and merge it with the `final_headers` variable. It will require running the serialization first and only then merging the headers (so explicitly passed ones have priority);

Similar changes may be needed for the WSGI integration, but let's start with `requests` first.

**Additional context**
Reported in [gitter](https://gitter.im/schemathesis/schemathesis?at=60908b6c2cc8c84d850b127f)

Stranger6667:
**TODO**:
- [x] Fix CLI to display tests on the same line
- [x] Write documentation
- [x] Fix #1152 (Internal errors on unusual schemas) as negative cases may not be objects
- [x] Implement constraint negation
- [x] Implement array schemas mutation (modify items value)
- [x] Handle schemas with multiple types
- [x] Consider caching JSON Schema validators & applying them instead of "allOf" + "not"
- [x] Implement guards on each mutation (so it mutates only appropriate types), maybe via a decorator
- [x] Make mutated properties required in `change_properties`
- [x] Apply swarm testing in `change_schema_type` , `remove_required_property` & `negative_schema` + maybe on the `location` level?
- [x] Add dependencies between mutations to avoid one mutation canceling another
- [x] Randomize ordering inside `get_mutations`
- [x] Do not generate data that can't be sent as `application/x-www-form-urlencoded`, e.g. `True` - the simplest option is to filter it later, but it is better to avoid such data by construction. Another way is to convert it to some valid data, e.g. cast to string & add to a dictionary.
- [x] Do not generate non-string values for headers / filter them out / cast to string
- [x] Do not generate empty header names
- [ ] Create a new check - if some required field was removed - and it is not 4xx, then the field is not really required. Might be worth reporting.
- [ ] Take the depth level into account when making location-dependent changes (some of them make sense only on the top level)
- [x] For multipart - it should always be generated (now it is not for `/multipart`)
- [x] Tests: fix `_aiohttp.form` handler failing on invalid bytes
- [x] Tests: fix `_aiohttp.create_user` failure on non-object type

Then existing tests should be adjusted/parametrized for negative scenarios.
Also, I plan to run it against the evaluation suite & test corpus. I am pretty sure that there will be quite a lot of things to work on after those runs and I'll add them here.

**Other considerations**:
- Path parameters are always cast to strings, and maybe there is no reason to change their type to, e.g., arrays/objects? Currently it is possible to have a path like this - `PATCH /api/users/%5B%5B%5B%5D%5D,%20%5B-12785%5D,%20%5B%5D,%20%7B%7D,%20%5B-12,%20''%5D,%20%7B'%5Cx9a%C2%A1`
- The current approach is more likely to mutate nodes that are closer to the root. I believe it will be better to make these changes more evenly distributed across the schema nodes. For, e.g., chose a subtree (+ apply swarm testing) & mutate it
- Detect whether some schema can be mutated at all before creating a strategy? E.g., run all possible mutations on it and exit on first successful. This way, it should be possible to detect whether it is possible to mutate a request at all and skip the test if it is not possible. Otherwise, it may lead to the "filter_too_much" health check failure & cost some performance. Health checks can be disabled, and then it can lead to an infinite execution.
- How `$ref` can affect the mutation process?
- Use JSON pointers in FeatureStrategy to skip nodes by their path, not by their name.

Resolves #65


src/schemathesis/specs/openapi/negative.py
```diff
@@ -0,0 +1,65 @@
...
+    if required:
+        required.remove(name)
+    # An optional property still can be generated
+    # To avoid it we need to remove it completely
+    properties = new_schema["properties"]
+    del properties[name]
+
+
+def mutate_type(draw: Draw, new_schema: Schema, property_schema: Schema, name: str) -> None:
+    to_exclude = {property_schema["type"]}
```
Zac-HD, line 55: The value of the `type` keyword may be a type *or a list of types*, in which case you may want to choose just one or exclude all (but then check that `available_types` is nonempty!).

src/schemathesis/specs/openapi/negative.py
```diff
@@ -0,0 +1,65 @@
...
+    if property_schema["type"] == "integer":
+        # Any valid integer is also a valid number
+        # If we won't change it then it is possible that the result will still match the original schema
+        # For example 0.0 is still a valid integer in JSON Schema Draft 7 (TODO. check)
```
Zac-HD, line 59: This is correct.  However it's *not* always the case in draft-04 IIRC, which is the one used by OpenAPI!

src/schemathesis/specs/openapi/negative.py
```diff
@@ -0,0 +1,65 @@
...
+        new_schema = deepcopy(schema)
+        properties = sorted(new_schema.get("properties", {}).items())
+        # At least one mutation should be done to at least one property
+        mutated_property_name, mutated_property_schema = draw(st.sampled_from(properties))
+        mutation = draw(st.sampled_from(MUTATIONS))
+        mutation(draw, new_schema, mutated_property_schema, mutated_property_name)
+
+        # Other properties & mutations are chosen with feature flags
+        enabled_properties = draw(FeatureStrategy())
+        enabled_mutations = draw(FeatureStrategy())
```
Zac-HD, line 31: So long as your names don't collide, you can use a single `FeatureStrategy` for both.

Unclear if you'd want to, but you could also use `st.shared(FeatureStrategy(), "key")` to have consistent feature exclusions across everything in the test case.

src/schemathesis/specs/openapi/negative.py
```diff
@@ -0,0 +1,65 @@
...
+Draw = Callable[[st.SearchStrategy[T]], T]
+Schema = Dict[str, Any]
+# draw, new_schema, property schema, name
+Mutation = Callable[[Draw, Schema, Schema, str], None]
+
+
+def negative_schema(schema: Dict[str, Any]) -> st.SearchStrategy:
+    """A strategy for objects, that DO NOT match the input schema."""
+
+    # First we mutate the input schema, so any valid instance for it is invalid to the original schema
```
Zac-HD, line 18: This makes sense; but note that one option which might be easier to implement in general (e.g. absence of a required `""` key when random extra properties are allowed) would be to calculate a negative schema with *little* overlap and then filter by "instance is not valid against original schema".

src/schemathesis/specs/openapi/negative.py
Stranger6667: Indeed! The empty `available_types` also leads to the case, when this mutation can't be applied at all. I am not 100% sure how to handle that, but there might be a case where there is no "negation" possible, e.g. with the `true` schema (and ones that canonicalise to it) - if the server is supposed to accept everything, we can't formally create anything that doesn't match the declared schema. Probably that schema can be rejected and marked as unsatisfiable - still, we can test all possible input ranges in the "positive" testing case and avoid overlapping.

src/schemathesis/specs/openapi/negative.py
Stranger6667: Good catch! I completely forgot about it! :) This makes this branch not relevant anymore (setting the right draft validator in tests confirms it)

src/schemathesis/specs/openapi/negative.py
Stranger6667: > So long as your names don't collide, you can use a single FeatureStrategy for both.

Do you mean checking it at runtime? I.e., if there are collisions, then create two strategies, otherwise create one for both?

> Unclear if you'd want to, but you could also use st.shared(FeatureStrategy(), "key") to have consistent feature exclusions across everything in the test case.

Makes sense! I probably missed the scope of Swarm testing initially - as far as I understand now, the feature exclusion should happen on the test case level, so some features remain disabled during the whole test case, and this basically what makes the Swarm testing work so well :)

src/schemathesis/specs/openapi/negative.py
Stranger6667: Interesting, I will try that! :)

src/schemathesis/specs/openapi/negative.py
Zac-HD: I mean "just create one, and it'll behave oddly-but-acceptably if a property is named `mutation_type` or whatever.

> Makes sense! I probably missed the scope of Swarm testing initially - as far as I understand now, the feature exclusion should happen on the test case level, so some features remain disabled during the whole test case, and this basically what makes the Swarm testing work so well :)

I don't know of experiments about swarm testing at various levels - heirarchal swarm testing seems intuitively reasonable (though would perform poorly with Hypothesis' current implementation), and global-per-test-case seems like the best option.

src/schemathesis/specs/openapi/negative.py
Stranger6667: I added `{"not": schema}` shortcut for schemas where the best way to negate is not clear - specific handling for such schemas can be added later. Also, now the output is filtered with `not validator.is_valid(s)`, which basically should work as you mention :) Then in mutations, there could be overlap in the resulting schema, but in the end, it will be filtered out :)

src/schemathesis/specs/openapi/negative.py
```diff
@@ -0,0 +1,115 @@
...
+        # In OAS 2 the "type" keyword is a list (in OAS 3 it is a string) therefore we need to convert "type" to a list
+        if "object" in get_type(new_schema):
+            return mutate_object_schema(draw, new_schema, mutations)
+        return {"not": new_schema}
+
+    return (
+        mutated()
+        .flatmap(lambda s: from_schema(s, custom_formats=custom_formats))
+        .filter(lambda i: not validator.is_valid(i))
+    )
```
Zac-HD, line 87:
```suggestion
    return mutated().flatmap(
        lambda s: from_schema({"allOf": [{"not": schema}, s]}, custom_formats=custom_formats)
    )
```

In principle `hypothesis-jsonschema` might be smarter about how to canonicalise this; in practice it will often fall back to exactly the same filter implementation in the worst case.

src/schemathesis/specs/openapi/_hypothesis.py
```diff
@@ -186,15 +190,17 @@ def is_yaml_parsing_issue(operation: APIOperation) -> bool:
...
+    parent_schema: BaseSchema,
 ) -> st.SearchStrategy:
     # The cache key relies on object ids, which means that the parameter should not be mutated
     # Note, the parent schema is not included as each parameter belong only to one schema
     if parameter in _BODY_STRATEGIES_CACHE and to_strategy in _BODY_STRATEGIES_CACHE[parameter]:
         return _BODY_STRATEGIES_CACHE[parameter][to_strategy]
     schema = parameter.as_json_schema()
     schema = parent_schema.prepare_schema(schema)
-    strategy = to_strategy(schema)
+    strategy = to_strategy(schema, "body")
```
buchi-busireddy, line 203: Should this be converted into an enum instead?

src/schemathesis/specs/openapi/negative/utils.py
```diff
@@ -0,0 +1,10 @@
+from typing import List
+
+from .types import Schema
+
+
+def get_type(schema: Schema) -> List[str]:
+    type_ = schema.get("type", ["null", "boolean", "integer", "number", "string", "array", "object"])
```
buchi-busireddy, line 7: Just curious, why `_` at the end of the variable name?

Zac-HD: Because "type" is the name of a Python built-in!

src/schemathesis/specs/openapi/negative/mutations.py
```diff
@@ -0,0 +1,153 @@
...
+    property_name = draw(st.sampled_from(sorted(required)))
+    required.remove(property_name)
+    if not required:
+        # In JSON Schema Draft 4, `required` must contain at least one string
+        # To keep the schema conformant, remove the `required` key completely
+        del schema["required"]
+    # An optional property still can be generated, and to avoid it, we need to remove it from other keywords.
+    properties = schema.get("properties", {})
+    properties.pop(property_name, None)
+    # This property still can be generated via `patternProperties`, but this implementation doesn't cover this case
```
buchi-busireddy, line 49: Just curious: Where can I see more details of patternProperties?

src/schemathesis/specs/openapi/negative/mutations.py
```diff
@@ -0,0 +1,153 @@
...
+    # The output filter covers cases like this
+    return MutationResult.SUCCESS
+
+
+def change_schema_type(draw: Draw, schema: Schema) -> MutationResult:
+    """Change type of values accepted by a schema.
+
+    Applicable types: any
+    """
+    if "type" not in schema:
```
buchi-busireddy, line 60: When would this happen?

src/schemathesis/specs/openapi/negative/mutations.py
```diff
@@ -0,0 +1,153 @@
...
+        # The absence of this keyword means that the schema values can be of any type;
+        # Therefore, we can't choose a different type
+        return MutationResult.FAILURE
+    types = set(get_type(schema))
+    candidates = {"string", "integer", "number", "object", "array", "boolean", "null"} - types
+    if not candidates:
```
buchi-busireddy, line 66: It seems like this code is checking the types of all top-level JSON keys together but there could be a lot of tests generated by changing the type of each JSON key separately right?

src/schemathesis/specs/openapi/negative/mutations.py
```diff
@@ -0,0 +1,153 @@
...
+    Mutations are sorted by their anticipated effectiveness. The ordering is essential when we need to unconditionally
+    apply at least one mutation.
+    """
+    types = get_type(schema)
+    # On the top-level of Open API schemas, types are always strings, but inside "schema" objects, they are the same as
+    # in JSON Schema, where it could be either a string or an array of strings.
+    # TODO. How to handle multiple types? Maybe each mutation can have a guard that will return FAILURE on a type that
+    # is not applicable
+    if "object" in types:
+        return change_properties, negate_constraints, remove_required_property, change_schema_type, negate_schema
```
buchi-busireddy, line 151: On what basis this order is maintained?

src/schemathesis/specs/openapi/negative/mutations.py
Stranger6667: It turned out that underspecified schemas are pretty common, not saying that it happens too often, but it is a visible fraction. Usually, I see cases like `{"format": "date"}` or `{"properties": {...}}` which are perfectly valid JSON Schemas.

src/schemathesis/specs/openapi/negative/mutations.py
Stranger6667: Mostly heuristic - from the most effective to the least effective. I still want to play & evaluate it, but the last mutation is negating everything, so it could be type, other constraints, etc., which seems like the least effective one, as it has the most broad spectrum of possible differences. I assume that smaller changes in the structure are better.

Zac-HD: > I assume that smaller changes in the structure are better.

I'd guess that "small" is the most effective single size, but that using a variety of sized changes finds even more bugs.

Stranger6667: > using a variety of sized changes finds even more bugs.

Could you, please, elaborate?

Zac-HD: Let's say that two endpoints allow invalid inputs of the correct type, and one allows invalid inputs only of the wrong type.

If we only make small changes (ie keep the type) we'll find two bugs, but if we sometimes keep the type and sometimes discard it we can find all three bugs.  ['Aim for maximum expressiveness in test cases'](https://blog.regehr.org/archives/591), even if we don't have full swarm-testing.

Stranger6667: Got it, thanks! 

I thought that this order is important only in the first step inside `change_properties` - when it is trying to apply at least one mutation to at least one property (to avoid cases when nothing is chosen by `FeatureStrategy` & it may be equivalent to the "positive" scenario). Other mutations to other properties are still possible in the second part of `change_properties`, depending on `FeatureStrategy`.

But now I see that in some cases, when there is only one property, then the latest mutation here might never be applied at all. E.g., for `{"foo": {"type": "string"}}`, it will always stop on `change_schema_type`. Probably randomized ordering might solve it, but maybe it could be weighted towards having `negate_schema` less probable?

Also, I completely forgot that applying `negate_schema` after others might cancel previous effects & lead to "positive" scenarios. E.g. `{"const": 42"}` will be negated twice, which gets us back to the same schema (however, it is the case with all schemas). So, probably some dependencies between mutations are needed.

src/schemathesis/specs/openapi/negative/mutations.py
Stranger6667: I think that the [JSON Schema pages](https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties) can provide a good overview :)

src/schemathesis/specs/openapi/negative/mutations.py
Stranger6667: Yep, those options are possible through `change_properties` that can apply other mutations including this one

src/schemathesis/specs/openapi/_hypothesis.py
Stranger6667: Could you, please, elaborate?

Zac-HD:
🎉 

Congratulations!  So exciting to have such a long-awaited feature 😁

Stranger6667:
Thank you! :) There will be some follow-up improvements (mostly in terms of data variety & performance), but I am happy to have it merged finally! :)

PR #853, opened 2020-11-08
fix: Handling of non-json content types

### LINKED ISSUE #850: [BUG] requestBody with text/plain gets wrapped in quotes when sent to server
**Describe the bug**

When specifying a request body as just a plain string, the request that is sent to the server is wrapped with double-quotes rather than the raw string being sent.

**To Reproduce**

1. Use the spec below
1. Hit a simple server endpoint that just dumps the request body (mine is in rust, so not that helpful)
1. Observe the data being wrapped in quotes.

```yaml
---
openapi: 3.0.0
info:
  title: Service
  version: '1.0.0'
paths:
  '/v1/parse':
    post:
      operationId: ParsePipeline
      requestBody:
        description: Pipeline syntax
        content:
          text/plain:
            schema:
              type: string
            examples:
              success:
                summary: 'Legit'
                value: 'hello world'

      responses:
        '200':
          description: Successfully parsed
```

**Expected behavior**

The raw string would be what I expected for sure :smile_cat:

 - OS:  Linux
 - Python version: 3.8.5
 - Schemathesis version: 2.7.1

**Additional context**

It looks like the use of `is_multlpart` in the `Case` class is the culprit here, since a plain string is viewed as "not" multipart and as such passed to `requests` as a `json` payload, rather than a `data` keyword argument.

I'm not sure what the right solution here is however. Monkey-patching such that a `str` body type is sent to requests as `data` definitely solves the problem but I'm sure that will have unintended consequences for other data types.

FWIW the swagger editor takes the above spec and sends the raw `hello world` string along to the server as I would expect

### LINKED ISSUE #852: [BUG] OpenApi30.get_hypothesis conversions doesn't handle `requestBody`
**Describe the bug**
In Open API 3 body is not in `parameters` and is not handled during the serialization phase.

Stranger6667:
I need to fix #852 first and then rework serialization.

Fixes #850

### DIFF
--- a/src/schemathesis/models.py
+++ b/src/schemathesis/models.py
@@ -188,21 +188,22 @@ class Case:  # pylint: disable=too-many-public-methods
         formatted_path = self.formatted_path.lstrip("/")  # pragma: no mutate
         url = urljoin(base_url + "/", formatted_path)
         # Form data and body are mutually exclusive
-        extra: Dict[str, Optional[Body]]
-        if self.form_data:
+        payload: Dict[str, Optional[Body]]
+        if self.form_data is not None:
             files, data = self.endpoint.prepare_multipart(self.form_data)
-            extra = {"files": files, "data": data}
-        elif is_multipart(self.body):
-            extra = {"data": self.body}
+            payload = {"files": files, "data": data}
+        elif self.body is not None:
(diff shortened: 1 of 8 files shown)


Stranger6667:
Closing in favor of #862

PR #857, opened 2020-11-10
feat: Generate tests for all defined media types

### LINKED ISSUE #472: Ignoring media type for request body (Openapi 3)
or at least they are not included in the case object.

Im converting case objects to curl commands for easier debugging and there is never any header present.

Im using this inside endpoint/post
```
"requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/simulationsRequest"
              }
            }
          }, 
          "description": "Body to give an input of desired simulations and their parameters."
        }
```

### LINKED ISSUE #690: Generate tests for all media types specified
When there are multiple media types specified within the "content" field, e.g. `application/json` and `multipart/form-data`, then we need to generate 2 test cases. Currently, we take only the first media type and ignore others

Stranger6667:
For each defined media type, there will be a test. The obvious benefit is that Schemathesis will be able to test more. However, it also simplifies the serialization step (basically a `map` function applied to the relevant strategy). Inside `get_case_strategy`, there will be media type available, so we can choose how to serialize any media type + allow to register custom serializers. As the next step, it will be easy to implement serializers for `text/plain`.

Resolves #690 
Resolves #472 

TODO:
- [ ] In CLI, all tests for the same endpoint should be displayed in the same line;
- [ ] A way to register custom serializers;
- [ ] Display media type in the test node name in pytest;
- [ ] Move "body" serialization out of `get_hypothesis_conversions`
- [ ] It will be nice to perform **any** body serialization in the function passed to `map`, including multipart & forms, so `get_requests_kwargs`, etc. will be cleaner
- [ ] Rename `get_hypothesis_conversions`, it is more about serialization
- [ ] Serialize forms & json in serializers
- [ ] Fix stateful schemas
- [ ] Fix explicit examples handling
- [ ] Documentation
- [ ] Make some separate structures to handle body / other parameters? 
- [ ] Use `consumes` for Swagger 2.0

### DIFF
--- a/src/schemathesis/_hypothesis.py
+++ b/src/schemathesis/_hypothesis.py
@@ -23,6 +23,7 @@ def create_test(
     settings: Optional[hypothesis.settings] = None,
     seed: Optional[int] = None,
     data_generation_method: DataGenerationMethod = DataGenerationMethod.default(),
+    media_type: Optional[str] = None,
     _given_args: Tuple[GivenInput, ...] = (),
     _given_kwargs: Optional[Dict[str, GivenInput]] = None,
 ) -> Callable:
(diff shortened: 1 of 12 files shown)


Stranger6667:
Closed in favor of  #862

PR #862, opened 2020-11-13
feat: Support generating tests for all defined media types

### LINKED ISSUE #690: Generate tests for all media types specified

### LINKED ISSUE #850: [BUG] requestBody with text/plain gets wrapped in quotes when sent to server

### LINKED ISSUE #916: [BUG] Body is always generated, even if optional
**Describe the bug**
Data generation for parameters with `in=body`  in Open API 2.0 and `requestBody` in Open API 3.0 always ignores if the payload is optional. It still produces the right payload, but the situation when the body is optional is not covered in generated tests at all.

**To Reproduce**
1. Run tests on a schema with optional body
2. Observe no requests without payload

Open API 2.0

```json
{
        "swagger": "2.0",
        "info": {"title": "Sample API", "description": "API description in Markdown.", "version": "1.0.0"},
        "host": "api.example.com",
        "basePath": "/v1",
        "schemes": ["https"],
        "paths": {
            "/users": {
                "post": {
                    "parameters": [
                        {"in": "body", "name": "body", "schema": {"type": "string"}, "required": false}
                    ],
                    "responses": {"200": {"description": "OK"}},
                }
            }
        },
    }
```

Open API 3.0:

```json
{
        "openapi": "3.0.2",
        "info": {"title": "Test", "description": "Test", "version": "0.1.0"},
        "paths": {
            "/query": {
                "get": {
                    "requestBody": {
                        "description": "Test.",
                        "content": {
                            "application/json": {"schema": {"type": "string"}}
                        },
                        "required": false,
                    },
                    "responses": {"200": {"description": "OK"}},
                }
            }
        },
    }
```

**Expected behavior**
I expect Schemathesis generate requests without payload too.

The solution would be just adding `| st.just(None)` to the final strategy.

Stranger6667:
Fixes #850
Resolves #690 
Fixes #916


src/schemathesis/models.py
```diff
@@ -441,6 +462,18 @@ def full_path(self) -> str:
...
 
+    def add_parameter(self, parameter: P) -> None:
+        lookup_table = {
+            "path": self.path_parameters,
+            "header": self.headers,
+            "cookie": self.cookies,
+            "query": self.query,
+            "body": self.body,
+        }
+        if parameter.location in lookup_table:
```
Stranger6667, line 473: Not sure if this condition is needed. If the parameter has a wrong location, then it should fail fast. May be `assert` will be enough or some internal exception

src/schemathesis/specs/openapi/_hypothesis.py
```diff
@@ -176,29 +200,19 @@ def quote_all(parameters: Dict[str, Any]) -> Dict[str, Any]:
...
+    endpoint: Endpoint, context: HookContext, hooks: Optional[HookDispatcher], strategy: st.SearchStrategy, key: str
+) -> st.SearchStrategy:
+    strategy = __apply_hooks(context, GLOBAL_HOOK_DISPATCHER, strategy, key)
+    strategy = __apply_hooks(context, endpoint.schema.hooks, strategy, key)
+    if hooks is not None:
+        strategy = __apply_hooks(context, hooks, strategy, key)
+    return strategy
+
+
+def __apply_hooks(
```
Stranger6667, line 213: one `_` should be ok

src/schemathesis/specs/openapi/examples.py
```diff
@@ -75,9 +75,13 @@ def get_request_body_example_from_properties(endpoint_def: Dict[str, Any]) -> Di
 def get_static_parameters_from_example(endpoint: Endpoint) -> Dict[str, Any]:
     static_parameters = {}
     for name in PARAMETERS:
-        parameter = getattr(endpoint, name)
-        if parameter is not None and "example" in parameter:
-            static_parameters[name] = parameter["example"]
+        parameters = getattr(endpoint, name)
+        if name == "body" and parameters and parameters[0].example:
```
Stranger6667, line 79: Both branches can be unified on the "container" level. It will work if there will be a single instance, not a list

src/schemathesis/specs/openapi/examples.py
```diff
@@ -111,14 +115,7 @@ def get_strategies_from_examples(endpoint: Endpoint, examples_field: str = "exam
 
 
 def get_strategy(endpoint: Endpoint, static_parameters: Dict[str, Any]) -> SearchStrategy[Case]:
```
Stranger6667, line 117: This function can be removed completely

src/schemathesis/specs/openapi/links.py
```diff
@@ -58,57 +59,40 @@ def parse(self, case: Case, response: GenericResponse) -> ParsedData:
...
-        # For each response from the previous endpoint run we create a new JSON schema and collect all results
-        # Instead of unconditionally creating `oneOf` list we gather it separately to optimize
-        # the resulting schema later
-        variants: Dict[str, List[Dict[str, Any]]] = {
-            "path_parameters": [],
-            "query": [],
-            "headers": [],
-            "cookies": [],
-            "body": [],
+        templates = {
```
Stranger6667, line 62: Add clarifications why it is implemented this way

src/schemathesis/specs/openapi/parameters.py
```diff
@@ -0,0 +1,348 @@
...
+            "in": "query",
+            "name": "key",
+            "type": "string"
+            "example": "foo",   # This one
+        }
+        """
+        return self.definition.get(self.example_field)
+
+    @property
+    def named_examples(self) -> Dict[str, Any]:
```
Stranger6667, line 68: This property is not used now - but need to re-check if the named examples are properly used.

src/schemathesis/specs/openapi/parameters.py
```diff
@@ -0,0 +1,348 @@
...
+        return self.definition.get(self.examples_field, {})
+
+    @property
+    def _schema_example(self) -> Any:
+        """Example defined on the schema-level.
+
+        Open API 3.0:
+
+        {
+            "in": "query",  (only "body" is possible for Open API 2.0)
```
Stranger6667, line 79: What about `formData`?

src/schemathesis/specs/openapi/parameters.py
```diff
@@ -0,0 +1,348 @@
...
+class OpenAPI20Parameter(OpenAPIParameter):
+    """Open API 2.0 parameter.
+
+    https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject
+    """
+
+    example_field = "x-example"
+    examples_field = "x-examples"
+    nullable_field = "x-nullable"
+    # https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject
```
Stranger6667, line 127: Should it be https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject instead?

The list of keywords should be re-checked

src/schemathesis/specs/openapi/parameters.py
```diff
@@ -0,0 +1,348 @@
...
+class OpenAPI30Parameter(OpenAPIParameter):
+    """Open API 3.0 parameter.
+
+    https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#parameter-object
+    """
+
+    example_field = "example"
+    examples_field = "examples"
+    nullable_field = "nullable"
+    # https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#schema-object
```
Stranger6667, line 171: Needs clarification, that in contrast with Open API 2, the schema is separated here in the `schema` keyword

src/schemathesis/specs/openapi/parameters.py
```diff
@@ -0,0 +1,348 @@
...
+        return "body"
+
+
+@attr.s(slots=True)
+class OpenAPI20Body(OpenAPIBody, OpenAPI20Parameter):
+    """Open API 2.0 body variant."""
+
+    def as_json_schema(self) -> Dict[str, Any]:
+        """Convert body definition to JSON Schema."""
+        # `schema` is required in Open API 2.0 bodies.
```
Stranger6667, line 229: "bodies" - the description can be more precise. It is related to the `in` value

src/schemathesis/specs/openapi/parameters.py
```diff
@@ -0,0 +1,348 @@
...
+    def transform_keywords(self, schema: Dict[str, Any]) -> Dict[str, Any]:
+        definition = super().transform_keywords(schema)
+        if self.is_form:
+            definition.setdefault("type", "object")
+        return definition
+
+    @property
+    def is_form(self) -> bool:
+        """Whether this payload represent a form."""
+        return self.media_type in ("multipart/form-data", "application/x-www-form-urlencoded")
```
Stranger6667, line 259: Are there any other media types for this purpose? Also, it could be a module-level constant

test/specs/openapi/test_parameters.py
```diff
@@ -0,0 +1,32 @@
...
+            "test",
+        ),
+        (
+            OpenAPI30Parameter,
+            {"name": "id", "in": "query", "required": True, "schema": {"type": "string", "example": "test"}},
+            "test",
+        ),
+    ),
+)
+def test_examples(cls, schema, expected):
```
Stranger6667, line 31: It will be nice to check if the precedence works as expected (examples within `schema` and top-level)

test/test_hypothesis.py
```diff
@@ -24,22 +24,33 @@ def make_endpoint(schema, **kwargs) -> Endpoint:
 @pytest.mark.parametrize("name", sorted(PARAMETERS))
 @pytest.mark.filterwarnings("ignore:.*method is good for exploring strategies.*")
 def test_get_examples(name, swagger_20):
-    example = {"name": "John"}
+    if name == "body":
+        # TODO. explain why
```
Stranger6667, line 28: Yes, explain it

src/schemathesis/specs/openapi/_hypothesis.py
```diff
@@ -69,72 +70,97 @@ def is_valid_query(query: Dict[str, Any]) -> bool:
...
+    query: Any = NOT_SET,
+    body: Any = NOT_SET,
+) -> Any:
+    to_strategy = {DataGenerationMethod.positive: make_positive_strategy}[data_generation_method]
 
+    context = HookContext(endpoint)
 
-def prepare_headers_schema(value: Dict[str, Any]) -> Dict[str, Any]:
-    """Improve schemas for headers.
+    if path_parameters is NOT_SET:
```
Stranger6667, line 90: maybe refactor? Branches look similar to each other

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -464,12 +468,13 @@ def prepare_multipart(
         files, data = [], {}
         # If there is no content types specified for the request or "application/x-www-form-urlencoded" is specified
         # explicitly, then use it., but if "multipart/form-data" is specified, then use it
+        # TODO. It might use the new Parameter API
```
Stranger6667, line 471: Re-check if it makes sense

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -481,8 +486,11 @@ def prepare_multipart(
         return files or None, data or None
 
     def get_request_payload_content_types(self, endpoint: Endpoint) -> List[str]:
-        global_consumes = endpoint.schema.raw_schema.get("consumes", [])
-        consumes = endpoint.definition.resolved.get("consumes", [])
+        return self._get_consumes_for_endpoint(endpoint.definition.resolved)
```
Stranger6667, line 489: Document this  method

PR #913, opened 2020-12-11
Update documentation for running in docker

gukoff:
"stable" tag ceased to exist.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -480,12 +480,12 @@ Schemathesis CLI is also available as a docker image:
 
 .. code-block:: bash
 
-    docker run schemathesis/schemathesis:stable \
+    docker run schemathesis/schemathesis:latest \
         run http://api.com/schema.json
 
 To run it against the localhost server, add ``--network=host`` parameter:
 
 .. code-block:: bash
 
-    docker run --network="host" schemathesis/schemathesis:stable \
+    docker run --network="host" schemathesis/schemathesis:latest \
         run http://127.0.0.1/schema.json


Stranger6667:
Hi! Good catch! But probably it will be better to revive the tag itself. It can be done by adding `tag: stable` [here](https://github.com/schemathesis/schemathesis/blob/master/.github/workflows/release.yml#L26)

gukoff:
Yup, this definitely works too. Can do so instead.

Are you planning to assign `stable` tag conditionally in the future? Because it looks like `stable` becomes synonymous to `latest`, which is probably not your end goal :)

Stranger6667:
> Are you planning to assign :stable tag conditionally in the future?

[These jobs](https://github.com/schemathesis/schemathesis/blob/master/.github/workflows/release.yml#L5) run on `published`; therefore, they should point to the latest release. It is an oversight that the "registry" job creates the `latest` tag; it [was](https://github.com/schemathesis/schemathesis/commit/b32216e7ad58a759380b3eb944eaf3205280acdf#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L17) `stable` before. The `latest` tag is created via a different [job](https://github.com/schemathesis/schemathesis/blob/master/.github/workflows/master_update.yml).

Stranger6667:
I think that we need to add `tag: stable` to both jobs in [this file](https://github.com/schemathesis/schemathesis/blob/master/.github/workflows/release.yml) and all images will be tagged properly.

Just to re-check. @aexvir, is it possible to assign both, the version tag (like `2.8.4`) and `stable` within the same step? As far as I understand from the action doc, the tag is guessed based on the ref name

Stranger6667:
I wasn't sure if kaniko supports multiple destinations, so [I added](https://github.com/schemathesis/schemathesis/commit/69b4c735cad06f29bce86fbe821ee05a9ca33f07) another step with the `stable` tag. Will check it in the next release

Stranger6667:
It works :) cc @gukoff 

```
⇒  docker run schemathesis/schemathesis:stable --version
schemathesis, version 2.8.5
```

Stranger6667:
Thank you for bringing attention to this issue :)

PR #1032, opened 2021-02-10
fix: Return a default terminal size to prevent crashes on systems with zero-width terminals

tuckerwales:
CircleCI returns a zero-width terminal which was causing the schemathesis CLI to crash during jobs. Providing a default prevents crashing in these cases.

The default terminal size of 80w 24h was chosen because it seems to be a historically 'standard' terminal size.

Ref: (https://softwareengineering.stackexchange.com/questions/148754/why-is-24-lines-a-common-default-terminal-height)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 `Unreleased`_ - TBD
 -------------------
 
+**Fixed**
+
+- Return a default terminal size to prevent crashes on systems with zero-width terminals (some CI/CD servers).
+
 **Added**
 
 - Support for external examples via the ``externalValue`` keyword. `#884`_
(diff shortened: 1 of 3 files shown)


Stranger6667:
Amazing catch! Thank you @tuckerwales ! :)
I will merge it once CI is done & make a bugfix release

tuckerwales:
> Amazing catch! Thank you @tuckerwales ! :)
> I will merge it once CI is done & make a bugfix release

No problem 👍 

Thanks, we look forward to using schemathesis in our pipeline - it's great!

Stranger6667:
> Thanks, we look forward to using schemathesis in our pipeline - it's great!

I am delighted to hear that :)

The fix is released in `3.0.9`. 

Note, that currently, Schemathesis fails to build on Alpine due to the recent changes in the `cryptography` package. Let me know if it affects your use case and I'll prioritize fixing it (tracking issue #1033).

PR #1034, opened 2021-02-11
fix: Return a minimum terminal width to prevent crashes on systems with zero-width terminals

tuckerwales:
Hello again!

Sorry, the fix that I implemented yesterday didn't work because I misunderstood how the `fallback` parameter of `shutil.get_terminal_size` worked. That value is only used if a terminal doesn't report a terminal size, but some CI/CD providers (like ours) DO provide a terminal size, it's just 0.

This fix will guarantee that a minimum terminal width of 80 is returned and I have tested this on our CI provider.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 `Unreleased`_ - TBD
 -------------------
 
+**Fixed**
+
+- Prevent a small terminal width causing a crash (due to negative length used in an f-string) when printing percentage
+
 **Added**
 
 - Support for external examples via the ``externalValue`` keyword. `#884`_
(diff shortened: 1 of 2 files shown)


Stranger6667:
Hi @tuckerwales !

As far as I see it will force `80` even on terminal windows with 0 < width < 80. How the output will look like then? 
I actually have a use case for this :) When I record terminal output for some presentation I usually shrink the window quite significantly (<80) and record/copy the output

tuckerwales:
Hey @Stranger6667!

You're absolutely correct, I didn't think of that. I have just corrected the change to fix the root of the problem, which was a negative length being used in the f-string in `display_percentage`. 

Now the percentage will be printed as far to the right of the terminal as it can be without causing a crash.

Hopefully this is ok 😄

Stranger6667:
Great! I just checked the latest version - it works as intended :) Thank you very much for your contribution :)

tuckerwales:
> Great! I just checked the latest version - it works as intended :) Thank you very much for your contribution :)

No problem, sorry the first fix didn't work!

Stranger6667:
> sorry the first fix didn't work!

No worries, hopefully now the issue is solved :) I'll make a new release soon

PR #1071, opened 2021-03-22
fix: Output types for evaluation results of `$response.body` and `$request.body` runtime expressions

### LINKED ISSUE #1068: [BUG] Runtime expressions do not preserve type for `$response.body`
**Describe the bug**
When Open API links are used and the `$response.body` expression is used it returns the whole body as a string, however, when `$response.body#/foo` is used, then the return type will have the same type as `foo`.

It should preserve the original body type.

Ref: #1048

Stranger6667:
Fixes #1068


Stranger6667:
Seems like a GH bug - the commit was actually merged

PR #1101, opened 2021-03-31
use functioning fixup name in example: fast_api instead of fastapi

bronsen:
https://github.com/schemathesis/schemathesis/blob/master/src/schemathesis/fixups/__init__.py#L5 uses `fast_api` and not `fastapi`.

### Description
The commented out example used "fastapi", but using that raised a KeyError in fixups/__init__.py. 

The fixup module is named "fast_api", as is the key in ALL_FIXUPS, so it seemed more acceptable to adapt the documentation to the code.

### Checklist
- [-] Created tests which fail without the change (if possible)
- [?] All tests passing (did not check)
- [-] Added a changelog entry
- [+] Extended the README / documentation, if necessary

### DIFF
--- a/docs/compatibility.rst
+++ b/docs/compatibility.rst
@@ -23,6 +23,6 @@ To use it, you need to add this code before you load your schema with Schemathes
     # will install all available compatibility fixups.
     schemathesis.fixups.install()
     # You can provide a list of fixup names as the first argument
-    # schemathesis.fixups.install(["fastapi"])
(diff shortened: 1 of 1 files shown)


Stranger6667:
Hi! Thank you for reporting :) Good catch.
There is also one similar case in `docs/faq.rst`. Would you mind updating it as well?

bronsen:
hm ok I'll try again with the conventionalcommits thingie

PR #1205, opened 2021-06-07
feat: A way to stop the Schemathesis runner's event stream manually via `events.stop()`

### LINKED ISSUE #1202: [FEATURE] A way to manually stop Schemathesis runner
**Is your feature request related to a problem? Please describe.**
When the Schemathesis runner is used directly, there is no direct way to stop it and return the `Finished` event

**Describe the solution you'd like**
It will be nice to have the `stop()` method on the returned generator

Stranger6667:
Resolves #1202

### DIFF
--- a/src/schemathesis/runner/impl/core.py
+++ b/src/schemathesis/runner/impl/core.py
@@ -1,5 +1,6 @@
 # pylint: disable=too-many-statements,too-many-branches
 import logging
+import threading
 import time
 import uuid
 from contextlib import contextmanager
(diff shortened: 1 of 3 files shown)


src/schemathesis/runner/impl/core.py
```diff
@@ -60,23 +61,35 @@ class BaseRunner:
     stateful_recursion_limit: int = attr.ib(default=DEFAULT_STATEFUL_RECURSION_LIMIT)  # pragma: no mutate
     count_operations: bool = attr.ib(default=True)  # pragma: no mutate
 
-    def execute(self) -> Generator[events.ExecutionEvent, None, None]:
+    def execute(self) -> "EventStream":
```
buchi-busireddy, line 64: Just for my own learning: Is this some kind of syntax to tell the return type in Python? And, was this introduced newly or something? I've not seen this kind of syntax used much in earlier Python programs.

Stranger6667: >  Is this some kind of syntax to tell the return type in Python? 

Yep! It is used when the type is not defined yet

> And, was this introduced newly or something? I've not seen this kind of syntax used much in earlier Python programs.

Originally, it was introduced in Python 3.5 (released in Sep 2015) as a part of [PEP-484](https://www.python.org/dev/peps/pep-0484/#forward-references). It is a syntax for forward references that may be not that common generally. Also, `from __future__ import annotations` affects this and if this import is in the module scope, then forward references do not require to be strings.

E.g.:


```python
class Foo:
    def bar(self) -> "Foo":
        pass
```

vs.

```python
from __future__ import annotations

class Foo:
    def bar(self) -> Foo:
        pass
```

PR #1256, opened 2021-08-30
Filters on API operations

### LINKED ISSUE #703: Ability to select multiple specific endpoints
Currently, we have two filters - `endpoint` and `method`, but what if we want to test only `GET /api/users` and `POST /api/orders`? It could be somehow done with `method=["GET", "POST"]` and `endpoint=["/api/users", "/api/orders"]` but it will also select `GET /api/orders` which we might not want.

It would be nice if the `endpoint` option can accept a list of tuples, where each tuple is a pair of method and path, then it might be possible to select any specific endpoint for Open API.

**Implementation notes**

- Update the type for `BaseSchema.endpoint` to include such tuples. 
- It might be easier to create a separate "filter" function in `schemathesis.specs.openapi.filters` that will accept the full path, method, and endpoints in this format
- This function might be called in [this condition](https://github.com/schemathesis/schemathesis/blob/master/src/schemathesis/specs/openapi/schemas.py#L70) and internally will be a no-op if endpoints are passed in the regular format. 
- At the same time, the current `should_skip_endpoint` function might be a no-op for this new format. The reason for this complication is to avoid extra work of resolving methods and dispatching the "before_process_path" hook.

### LINKED ISSUE #819: [FEATURE] Exclude endpoints
**Is your feature request related to a problem? Please describe.**
It is complicated to exclude endpoints from testing. E.g. if you want to exclude one endpoint then you either need to use `pytest -k "not ..."` syntax, or some complex regular expression or compose the testing using only the "include" direction

**Describe the solution you'd like**
Maybe arguments that will exclude endpoints from testing. Not sure for now

### LINKED ISSUE #1006: [FEATURE] Filter based on custom properties x-*
**Is your feature request related to a problem? Please describe.**
I transform my API Description from `raml` to OpenAPI3 and therefore can not make the use out of tags.
The only way to mark operations with custom information is via the `vendor extension` (https://swagger.io/docs/specification/openapi-extensions/)

**Describe the solution you'd like**

It would be helpful if I could filter tested operation by a list of `vendor-extensions`.

`--vendor-extension x-test true --vendor-extension x-imaginary false`

Stranger6667:
Resolves #703 
Resolves #819
Resolves #1006 

TODO:
- [x] Deprecation docs
- [x] Deprecate passing filters to loaders
- [x] Deprecate passing filters to `parametrize`
- [ ] Use `include` in tests
- [ ] Deprecate using old CLI arguments for filtration
- [ ] Make CLI use `include` / `exclude`
- [ ] CLI support for custom filters. It could be some simple `jq`-like filter - `--include-by="x-foo == 42"`, where the expression is of form "JSONPath Operator Value"
- [ ] CLI support for new filters - `--include-method`, `--exclude-method`, etc.
- [ ] How to remove filters?
- [x] Implement `exclude`
- [ ] Add `include` / `exclude` for GraphQL schemas
- [ ] Add relevant methods on "lazy" schemas
- [ ] Docs (CLI & Python) & comments
- [x] Use `path` instead `endpoint`?
- [ ] Separate method for regular expressions?
- [x] How to deal with lists of strings?
- [x] Better repr for filter functions
- [x] Deal with filters overriding

In Schemathesis 4.0 it will be nice to remove those arguments & attributes

### DIFF
--- a/src/schemathesis/schemas.py
+++ b/src/schemathesis/schemas.py
@@ -266,11 +266,12 @@ class BaseSchema(Mapping):
         app: Any = NOT_SET,
         hooks: Union[HookDispatcher, NotSet] = NOT_SET,
         validate_schema: Union[bool, NotSet] = NOT_SET,
-        skip_deprecated_operations: Union[bool, NotSet] = NOT_SET,
+        skip_deprecated_operations: Union[Optional[bool], NotSet] = NOT_SET,
         data_generation_methods: Union[Iterable[DataGenerationMethod], NotSet] = NOT_SET,
         code_sample_style: Union[CodeSampleStyle, NotSet] = NOT_SET,
         filters: Union[List[BaseFilter], NotSet] = NOT_SET,
     ) -> S:
+        # pylint: disable=too-many-branches
         if base_url is NOT_SET:
             base_url = self.base_url
         if method is NOT_SET:
@@ -293,9 +294,26 @@ class BaseSchema(Mapping):
             data_generation_methods = self.data_generation_methods
         if code_sample_style is NOT_SET:
             code_sample_style = self.code_sample_style
-        new_filters = self._construct_filters(
-            method, endpoint, tag, operation_id, cast(bool, skip_deprecated_operations), Include
-        )
+        unique_filters = {
+            filter_.group_id: filter_
+            for filter_ in self._construct_filters(
+                method, endpoint, tag, operation_id, cast(bool, skip_deprecated_operations), Include
+            )
+        }
+        for filter_ in self.filters:
+            unique_filters.setdefault(filter_.group_id, filter_)
+        # TODO. should it be on None? `None` could come from `self` attribute
+        if endpoint is None:
+            unique_filters.pop(1, None)
+        if method is None:
+            unique_filters.pop(2, None)
+        if tag is None:
+            unique_filters.pop(3, None)
+        if operation_id is None:
+            unique_filters.pop(4, None)
+        if skip_deprecated_operations is None:
+            unique_filters.pop(5, None)
+        new_filters = list(unique_filters.values())
         if filters is not NOT_SET:
             new_filters += cast(List[BaseFilter], filters)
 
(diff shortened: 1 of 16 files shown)


src/schemathesis/schemas.py
```diff
@@ -358,6 +394,18 @@ def validate_response(self, operation: APIOperation, response: GenericResponse)
...
+    def include_by(self, predicate: Callable) -> "BaseSchema":
+        """Get a new schema that includes API operations that pass the given predicate."""
+        return self._filter_by(Include(predicate))
+
+    def exclude_by(self, predicate: Callable) -> "BaseSchema":
+        """Get a new schema that excludes API operations that pass the given predicate."""
+        return self._filter_by(Exclude(predicate))
+
+    def _filter_by(self, *predicates: BaseFilter) -> S:
+        filters = self.filters + list(predicates)
```
Stranger6667, line 406: Maybe have a way to combine filters via OR

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -112,6 +115,54 @@ def __repr__(self) -> str:
         info = self.raw_schema["info"]
         return f"{self.__class__.__name__} for {info['title']} ({info['version']})"
 
+    def include(
+        self,
+        path: Optional[Filter] = None,
```
Stranger6667, line 120: Having ‘method’ as the first argument would be better

src/schemathesis/schemas.py
```diff
@@ -203,6 +217,10 @@ def parametrize(
         _code_sample_style = (
             CodeSampleStyle.from_str(code_sample_style) if isinstance(code_sample_style, str) else code_sample_style
         )
+        for name in ("method", "endpoint", "tag", "operation_id", "skip_deprecated_operations"):
+            value = locals()[name]
+            if value is not NOT_SET:
+                warn_filtration_arguments(name)
```
Stranger6667, line 223: Use existing filters, then replace / add new ones from these values. Then pass the result to clone

Stranger6667: In this case, clone will be simple

src/schemathesis/schemas.py
```diff
@@ -274,8 +294,30 @@ def clone(
             data_generation_methods = self.data_generation_methods
         if code_sample_style is NOT_SET:
             code_sample_style = self.code_sample_style
-
-        return self.__class__(
+        unique_filters = {
```
Stranger6667, line 297: This part should be the other way around - it will be simpler to remove filters

PR #1295, opened 2021-10-06
Feature/966 refresh auth

JangasCodingplace:
🚨 DRAFT 🚨
Before I finalise everything and diving deeper into existing project logic, please do a high level review for coding style and feature solving.

Black and Linting are not included yet.
It should be just a first review through pure functionality and logical project anti pattern if possible.

### Description

### Checklist
- [ ] All tests passing
- [ ] Added a changelog entry

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,3 @@
 [pytest]
 norecursedirs = .hypothesis .idea .git src docs .pytest_cache .mypy_cache .tox
+addopts = --auth janis:superpwd
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,2 @@
 [pytest]
 norecursedirs = .hypothesis .idea .git src docs .pytest_cache .mypy_cache .tox
-addopts = --auth janis:superpwd
--- a/src/schemathesis/auth.py
+++ b/src/schemathesis/auth.py
@@ -1,0 +1,93 @@
+import requests
+from enum import Enum
+from typing import Optional, Any
+from functools import partial
+from requests.auth import HTTPBasicAuth
+
+
+class AuthType(Enum):
+    COOKIE = 0
+    HEADER = 1
+    QUERY_PARAMS = 2
+    USERNAME_PASSWORD = 3
+
+
+class AuthStorage:
+    """Description
+
+    Szenarios:
+        No Token/Auth Set for request yet
+        Token Expires
+
+    Auth Methods
+        Cookie
+        Header
+        request auth
+
+    Keyword Arguments:
+        username
+        password
+        auth_token
+    """
+
+    login_url: Optional[str] = None
+    refresh_url: Optional[str] = None
+    auth_type: AuthType = AuthType.HEADER
+    auth_callback: Optional[Any] = None
+
+    DEFAULT_CONFIG = {
+        "username": None,
+        "password": None,
+        "token": None
+    }
+
+    def __init__(self, **configs):
+        if self.auth_type == AuthType.HEADER:
+            pass
+        elif self.auth_type == AuthType.COOKIE:
+            pass
+        elif self.auth_type == AuthType.QUERY_PARAMS:
+            pass
+        elif self.auth_type == AuthType.USERNAME_PASSWORD:
+            self._username = configs.get("username")
+            self._password = configs.get("password")
+        else:
+            raise TypeError("auth_type not supported.")
+
+    def _perform_username_password_authentication(self):
+        """Global override of request.request method with auth callback
+
+        Requirement: Each authentication needs to be valid on adding username / password combination
+            per request
+        """
+        if not self.auth_callback:
+            # Warning: Overriding static param will have singleton side effect
+            self.auth_callback = HTTPBasicAuth
+        requests.request = partial(
+            requests.request,
+            auth=self.auth_callback(
+                username=self._username,
+                password=self._password
+            )
+        )
+
+    def perform_authentication(self):
+        if self.auth_type == AuthType.HEADER:
+            pass
+        elif self.auth_type == AuthType.COOKIE:
+            pass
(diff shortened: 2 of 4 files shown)


JangasCodingplace:
```
@schemathesis.register_auth_provider(refresh_interval=600)
def authorize(auth):
    response = requests.post(URL, data=data)
    token = response.json()["access_token"]
    auth.set_header("Authorization", f"Bearer {token}")
```

My Job: Write the `register_auth_provider` which will be executed __always__ and refresh the token if `now - last_execution > refresh_interval`

Users Job: write the `authorize` method

Correct?

--

If so, we do have another difference in behaviour expectation:
My intention was to use that decorator at any method which should use an authorization. My experience: Some endpoints are not accessable if you are authenticated.

I was expecting something like the @login_required decorator which I know from django already and want to explicite declarate a method which requires an authentication with that @auth.register_auth_provider decorator.

Stranger6667:
### Usage workflows

On the top level there are two main workflows in Schemathesis:
1. Python tests. It uses the `pytest` runner. Users have to write their own tests, everything is integrated with the pytest machinery (parametrization, fixtures, etc). If the user wants to extend a test, then it is more flexible. People usually use it when they have existing Python projects with pytest-based test suites.
2. CLI. Implements its own test runner. Users do not write tests - the runner provides [some templates](https://github.com/schemathesis/schemathesis/blob/master/src/schemathesis/runner/impl/core.py#L520) that are not directly accessible, but extendable via [hooks](https://schemathesis.readthedocs.io/en/stable/extending.html#hooks). People use it with non-Python projects or when they don't need much flexibility, as it is much simpler to start with.

Both these approaches have a slightly different set of features - for example, they utilize different approaches for stateful testing. But in the future, I'd like to have feature parity between them.

I'd like to note that the approaches are quite different at the implementation level. `pytest` has its own goals and architecture - sometimes it is harder to implement features for the pytest runner because of it. On the other hand, the CLI runner is completely self-contained and it is quite easy to extend & implement new features. Its output style was directly inspired by `pytest`.

This particular feature **should** be implemented for both approaches but **might** be implemented as separate steps.

### General goals

I think that the top priority is the simplicity of usage - the fewer actions the end-user will do, the better. Of course, at the same time, it should not limit the user - they should be able to implement any auth flow.

Another thing is that it will be nice to have the same user-facing API for both, Python tests & CLI. Generally, with other hooks, users are able to solve their task within a single function (e.g. adjust data generation of headers). So, ideally, it might be the same decorator for both use cases.

### Pytest details

Usually, everything goes through a single test:

```python
@schema.parametrize()
def test_api(case):
    case.call_and_validate()
```

It tests all API operations available in the schema - it is possible to select a subset of them though. The way the execution flow goes is roughly this:
- `pytest` [discovers](https://github.com/schemathesis/schemathesis/blob/master/src/schemathesis/extra/pytest_plugin.py#L198) the `test_api` function. It contains a special marker added by the `schema.parametrize` decorator.
- Then it [expands](https://github.com/schemathesis/schemathesis/blob/master/src/schemathesis/extra/pytest_plugin.py#L187) this test into many tests, similar to what `pytest.mark.parametrize` is doing - separate test for each API operation.
- Each generated test is using the `test_api` function body which is [wrapped](https://github.com/schemathesis/schemathesis/blob/78dd61729e0eaa5b03249aa55409553b919246c4/src/schemathesis/_hypothesis.py#L19) with Hypothesis's machinery - e.g. the `given` decorator. Which makes each generated test be executed 100 times (by default) with different `case` values.
- Each generated test is [executed](https://github.com/schemathesis/schemathesis/blob/master/src/schemathesis/extra/pytest_plugin.py#L211) by `pytest` - it has some additional error handling.

In the end, `test_api` is executed up to M * N times, where M is the number of API operations to be tested and N is the value of Hypothesis's `max_examples` config option.

The `case` generation is [based](https://github.com/schemathesis/schemathesis/blob/78dd61729e0eaa5b03249aa55409553b919246c4/src/schemathesis/specs/openapi/_hypothesis.py#L100) on Hypothesis strategies, where the `Case` instance is constructed from different parts.

To modify the `case` instances I see at least two places:
- Inside the Hypothesis strategy - though, it might be a bit problematic, as strategies do not like external state (it might lead to `Flaky` failures, but I am not yet sure what would be the real effect of this)
- Inside the test body - here the user will need to call some function / method to adjust `case` with proper auth info.

Most likely I overestimate the importance of the "Flaky" case (it might happen with the "test body" option as well), so we might try to go with the former option.

Another important part is how to pass the registered function to the strategy. I think if we go with `schema.register_auth_provider`, then it can store the function on the `schema` instance, which is available during test [generation](https://github.com/schemathesis/schemathesis/blob/78dd61729e0eaa5b03249aa55409553b919246c4/src/schemathesis/extra/pytest_plugin.py#L85) and [pass](https://github.com/schemathesis/schemathesis/blob/78dd61729e0eaa5b03249aa55409553b919246c4/src/schemathesis/extra/pytest_plugin.py#L107) it to each generated test.

### CLI details

CLI is implemented on top of the [runner](https://github.com/schemathesis/schemathesis/tree/master/src/schemathesis/runner) and uses the same Hypothesis strategies & Hypothesis tests as pytest. Though, it executes them differently - all the failure handling is [implemented](https://github.com/schemathesis/schemathesis/blob/master/src/schemathesis/runner/impl/core.py#L225) in-house. It is based around an event stream model - the whole flow is a generator that emits events that denote certain steps of the testing process (e.g. before test, after test).

The implementation should be quite similar to `pytest` as if we [put](https://github.com/schemathesis/schemathesis/blob/78dd61729e0eaa5b03249aa55409553b919246c4/src/schemathesis/runner/impl/solo.py#L31) it into a strategy, then it will "just" work here as well.


P.S. While I writing this, I realize, that the scope of changes is much bigger than I initially anticipated

Stranger6667:
Closing in favor of #1462

PR #1371, opened 2022-01-13
fix: Generating illegal unicode surrogates in queries #1370

### LINKED ISSUE #1370: [BUG] Avoid generating illegal surrogates for list query parameters
**Describe the bug**
There could be an error like this: `UnicodeEncodeError: 'utf-8' codec can't encode character '\udf06' in position 17: surrogates not allowed`.

**To Reproduce**
OpenApi schema to reproduce:

```yaml
      schema:
        type: array
        items:
          type: string
```
or
```yaml
      schema:
        type: array
        items:
          type: string
          pattern: "^[^\\x00]*$"
```

**Expected behavior**
Avoid illegal surrogates generation for the described case

(generally unimportant)
- OS - does not matter
- python - does not matter
- Schemathesis version 3.12.2
- Spec version 3.0.0

gluhar2006:
fixes #1370

### DIFF
--- a/src/schemathesis/specs/openapi/_hypothesis.py
+++ b/src/schemathesis/specs/openapi/_hypothesis.py
@@ -87,7 +87,7 @@ def is_illegal_surrogate(item: Any) -> bool:
         return isinstance(value, str) and bool(re.search(r"[\ud800-\udfff]", value))
 
     if isinstance(item, list):
-        return any([check(item_) for item_ in item])
+        return any(check(item_) for item_ in item)
     return check(item)
 
 
--- a/src/schemathesis/specs/openapi/_hypothesis.py
+++ b/src/schemathesis/specs/openapi/_hypothesis.py
@@ -83,7 +83,12 @@ def is_valid_header(headers: Dict[str, Any]) -> bool:
 
 
 def is_illegal_surrogate(item: Any) -> bool:
-    return isinstance(item, str) and bool(re.search(r"[\ud800-\udfff]", item))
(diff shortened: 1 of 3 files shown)


src/schemathesis/specs/openapi/_hypothesis.py
```diff
@@ -83,7 +83,12 @@ def is_valid_header(headers: Dict[str, Any]) -> bool:
 
 
 def is_illegal_surrogate(item: Any) -> bool:
-    return isinstance(item, str) and bool(re.search(r"[\ud800-\udfff]", item))
+    def check(value: Any) -> bool:
+        return isinstance(value, str) and bool(re.search(r"[\ud800-\udfff]", value))
+
+    if isinstance(item, list):
+        return any([check(item_) for item_ in item])
```
Stranger6667, line 90: This one should be sufficient (without an intermediate list)
```suggestion
        return any(check(item_) for item_ in item)
```

Stranger6667:
Btw, I disabled the `duplicate-code` lint, so it should not interfere with the actual issue anymore

Stranger6667 (changes requested):
The abovementioned change is required to pass `pylint`, otherwise we're good to go!

src/schemathesis/specs/openapi/_hypothesis.py
gluhar2006: Thank you. I was too lazy to install pre-commit, my bad

Stranger6667 (approved):

PR #1380, opened 2022-02-04
Update example docker files

niya3:
### Description

Simplified and stabilized docker-related stuff in `example` dir:

1. removed unused deb-packages which were needed to build some old `asyng` wheels
2. binded web app port to `127.0.0.1` instead of `0.0.0.0` , it's more secure
3. added option to recreate web app container on failure, it helps to start web app in robust way
4. fixed depended docker image version to `postgres:14`, the current one
5. mounted `main.py` to containter (run-time dependency) instead of copying it (build-time dependency)


Stranger6667:
Looks good to me! Thanks :)

PR #1411, opened 2022-03-29
dependency: change werkzeug version range to < 2.1.0 to fix error

Bidaya0:

### Description
change werkzeug version range to < 2.1.0 to fix ModuleNotFoundError

### Checklist
- [ ] All tests passing
- [ ] Added a changelog entry

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ pytest-subtests = ">=0.2.1,<1.0"
 requests = "^2.22"
 click = "^8.0"
 importlib_metadata = { version = ">=1.1,!=3.8", python = "<3.8" }
-werkzeug = ">=0.16.0"
+werkzeug = ">=0.16.0,<2.1.0"
 junit-xml = "^1.9"
 starlette = ">=0.13,<1"
 yarl = "^1.5"


Stranger6667:
Hi!
Thank you for providing a patch! Though, it might be better to support newer Werkzeug versions instead - some conditional logic with creating an empty mixin if Werkzeug is before 2.1.0

Stranger6667:
@Bidaya0 

What do you think?

briantownley-wpe:
Seems like moving to a newer version will take longer (due to the changes in the package), but there are people like me who's production builds are broken due to this issue. 

Could we get the bug resolved in this PR first to unblock people and then have a new PR to handle upgrading to the newer version?

Stranger6667:
@briantownley-wpe 

Thanks for pinging me. I'll work on resolving this now

Bidaya0:
> @Bidaya0
> 
> What do you think?

I hold the same point of view as @briantownley-wpe , not only because I have the same experience with him (in the CICD process, due to the schemathesis error, the normal CICD process cannot be followed).
I've found other dependencies when updating dependencies that are treated the same way as this one, which may not be a sensible release measure, especially when the python package's dependencies are not version locked with release.
The project is still at risk, as this time when a breaking change in a dependency affects schemathesis

Stranger6667:
Will release `3.13.4` once #1415 jobs are finished (the `black` failure is unrelated to the change)

PR #1456, opened 2022-04-16
feat: Make the `SCHEMA` CLI argument accept API slug from Schemathesis.io

Stranger6667:
The main idea is to enable people to run `st run slug` where "slug" is a shortcut name for an API in Schemathesis.io, so the actual schema address could be loaded from there.

It would be nice to avoid calling the service for API config before validating the "api_slug" argument - maybe assume that if `SCHEMA` is not an URL / path, then it is always a slug, then just throw an error if `api_slug` is present?

TODO:
- [ ] Update the docs
- [ ] docstrings
- [x] Write the tests
- [x] Validate `api_slug`
- [x] Load config from Schemathesis.io & merge it

### DIFF
--- a/src/schemathesis/cli/callbacks.py
+++ b/src/schemathesis/cli/callbacks.py
@@ -1,3 +1,4 @@
+import enum
 import os
 import re
 from contextlib import contextmanager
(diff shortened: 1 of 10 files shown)


src/schemathesis/cli/callbacks.py
```diff
@@ -13,28 +14,39 @@
...
-            if "\x00" in raw_value or not utils.file_exists(raw_value):
-                raise click.UsageError("Invalid SCHEMA, must be a valid URL or file path.")
-            if "base_url" not in ctx.params and not ctx.params.get("dry_run", False):
-                raise click.UsageError('Missing argument, "--base-url" is required for SCHEMA specified by file.')
-        else:
-            _validate_url(raw_value)
-    return raw_value
+
+@enum.unique
+class SchemaVariant(enum.Enum):
```
Stranger6667, line 21: Maybe move to the CLI root instead?

PR #1516, opened 2022-05-19
feat: Report uploading to Schemathesis.io

Stranger6667:
TODO:
- [ ] Command help
- [ ] Forbid using `--report` when the regular integration is enabled
- [ ] Report message
- [ ] Parse the service response
- [ ] Comments & docstrings
- [ ] Simplify branching code
- [ ] Support saving report as a file / sending to another endpoint

### DIFF
--- a/src/schemathesis/cli/__init__.py
+++ b/src/schemathesis/cli/__init__.py
@@ -391,6 +391,7 @@ with_hosts_file = click.option(
 @click.option(
     "--junit-xml", help="Create junit-xml style report file at given path.", type=click.File("w", encoding="utf-8")
 )
+@click.option("--report", is_flag=True)
 @click.option(
     "--debug-output-file",
     help="Save debug output as JSON lines in the given file.",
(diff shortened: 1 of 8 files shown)


src/schemathesis/cli/__init__.py
```diff
@@ -640,7 +642,8 @@ def run(
                 base_url = base_url or test_run.config.base_url
         except requests.HTTPError as exc:
             handle_service_error(exc)
-
+    if report and client is None:
+        client = service.ServiceClient(base_url=schemathesis_io_url, token=None)
```
Stranger6667, line 646: Is it needed at all? there is no token, so it could be just a single `requests` call with the proper User-Agent.

src/schemathesis/cli/__init__.py
```diff
@@ -973,15 +977,20 @@ def execute(
...
-    if client is not None and test_run is not None:
+    if client is not None:
         service_queue: Queue = Queue()
         service_context = ServiceContext(url=schemathesis_io_url, queue=service_queue)
-        reporter = service.ServiceReporter(client=client, test_run=test_run, out_queue=service_queue)
-        handlers.append(reporter)
+        if test_run is not None:
+            reporter = service.ServiceReporter(client=client, test_run=test_run, out_queue=service_queue)
+            handlers.append(reporter)
+        if report:
```
Stranger6667, line 992: handle it separately? not within this branch

PR #1530, opened 2022-06-23
fix: Do not discard dots in OpenAPI expressions during parsing

frederikaalund:
### Description

Allow dots in OpenAPI expressions outside variable definitions. Currently, the parser discards these dots. With this PR, the parser outputs dots as strings.

#### Motivating example

[Postgrest horizontal filtering via query parameters](https://postgrest.org/en/stable/api.html#horizontal-filtering-rows):
```
GET /people?age=lt.13
GET /people?age=gte.18&student=is.true
GET /users?id=eq.42
```
OpenAPI links to these operations require the dot. E.g., `eq.{$response.body#/0/id}`. Currently, schemathesis outputs `id=eq42` that fails because of the missing dot. With this PR, the same expression outputs `id=eq.42` that works.

### Checklist
- [Y] Created tests which fail without the change (if possible)
- [Y] All tests passing
- [Y] Added a changelog entry
- [N/A] Extended the README / documentation, if necessary

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 `Unreleased`_ - TBD
 -------------------
 
+**Fixed**
+
+- Do not discard dots (``.``) in OpenAPI expressions during parsing.
+
 .. _v3.15.5:
 
 `3.15.5`_ - 2022-06-21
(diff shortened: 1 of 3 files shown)


Stranger6667:
Hi @frederikaalund 

Thank you very much for the detailed report & patch! :) Makes perfect sense to me

Stranger6667:
Released in `3.15.6`

PR #1563, opened 2022-07-19
fix: failure on response of apiserver in utf-8 encoding with BOM

egorlutohin:
### Description
If api server response in utf-8 and contain BOM, cli fails with exception message "The received response is not valid JSON: ... Exception: Unexpected UTF-8 BOM (decode using utf-8-sig): line 1 column 1 (char 0)". 

Answer in utf-8 with BOM is default behaviour in wide-used (in small areas) 1C:Enterprise platform (https://1c-dn.com/1c_enterprise/). 1C:Enterprise platform have tricky workarounds to switch off BOM, but if schemathesis will includes this fix by default, it will be better.

In this patch applyed knowledge from [How to fix Unexpected UTF-8 BOM error when using json.loads](https://www.howtosolutions.net/2019/04/python-fixing-unexpected-utf-8-bom-error-when-loading-json-data/) article.

### Checklist
- [x] All tests passing
- [x] Added a changelog entry


src/schemathesis/specs/openapi/schemas.py
```diff
@@ -479,6 +479,10 @@ def validate_response(self, operation: APIOperation, response: GenericResponse)
             return
         try:
             if isinstance(response, requests.Response):
+
+                if response.encoding == 'utf-8' and (response.text[0:1] in ('\ufeff', '\ufffe')): # utf-8 with BOM
```
Stranger6667, line 483: Could we have the byte order mark char as module-level constant instead? Btw, why there are two of them?

Also, a comment about the motivation would be nice. E.g. that RFC 7159 [forbids](https://www.rfc-editor.org/rfc/rfc7159#section-8.1) BOM, but we need this to workaround some non-compliant servers.

Stranger6667 (changes requested):
Hi @egorlutohin

Thank you for the patch! Generally, I'd prefer this to be solved on `requests` level (like [this](https://github.com/psf/requests/pull/4976/files) PR, which was reverted a while back), but I am fine with this approach. 
I have a minor comment on the implementation & it could be merged. I'll write a test for it later.

I don't want to complicate this PR now, but as the next step I'd move it to fixups (like with `fastapi`), but there is no hook in Schemathesis to build a fixup for this, so I need to implement such a hook first. The main motivation behind this is that Schemathesis tries to be compliant with standards when it is possible and provide such optional fixups for cases when they are needed.

src/schemathesis/specs/openapi/schemas.py
egorlutohin: Thank you for your attention to this question and interesting links about it!

> Could we have the byte order mark char as module-level constant instead? 
Of course!

> Btw, why there are two of them?
As mentioned in [wikipedia BOM article](https://en.wikipedia.org/wiki/Byte_order_mark):

```
 * If the 16-bit units are represented in [big-endian](https://en.wikipedia.org/wiki/Big-endian) byte order, the BOM will appear in the sequence of bytes as **0xFE 0xFF**
  * If the 16-bit units use [little-endian](https://en.wikipedia.org/wiki/Little-endian) order, the BOM will appear in the sequence of bytes as **0xFF 0xFE**
```
On my intel x64 box those bytes appears as '\ufeff', but maybe on other platforms it might be '\ufffe' in that situation, therefore I'd add this.

Approach proposed in `requests` PR (use to all repsonses in utf-8 python's utf-8-sig encoding to decode) also works good  in my situation, I have tested. As I read after on [python codecs docs](https://docs.python.org/3/library/codecs.html#encodings-and-unicode): "On decoding utf-8-sig will skip those three bytes if they appear as the first three bytes in the file.", but if utf-8 have no BOM, 'utf-8-sig' acts linke 'utf-8' encoding, I can confirm it with my tests :)

I trust you to choose best solution! Thank you in advance!

Stranger6667: Hmm, the article section you are referring to is about UTF-16, right? Here we have UTF-8, but I am not sure if `\ufffe` is relevant to it

Stranger6667: Though, it could be tested :) I'll check it out

egorlutohin: > Hmm, the article section you are referring to is about UTF-16, right? Here we have UTF-8, but I am not sure if `\ufffe` is relevant to it

Well, as I'd investigated, BOM in UTF-8 is 0xEF 0xBB 0xBF bytes, and that BOM not about byte order, it's about signaling that it's UTF-8 data (it so happened). But, when `requests` receive raw http output, it converts received bytes to internal python string representation through decoding raw data with 'utf-8' encoder. Python internal representation according to [PEP 393](https://peps.python.org/pep-0393/) may be LATIN1, UCS2 (UTF-16) or UCS4 (UTF-32).

In my case python's internal representantion was USC2/UTF-16, thus on my box 0xEF 0xBB 0xBF bytes of UTF-8 becomes `\ufffe` in python's string.

As far as I dive in this, `requests` PR solution to use 'utf-8-sig' to decode 'utf-8' seems be better solution. Sorry if I confused at all.

Stranger6667:
Hi @egorlutohin! In the next release, Schemathesis will handle BOM if `--fixups=utm8_bom` is used

PR #1576, opened 2022-08-03
fix: verbosity help message

lina-is-here:
### Description
I've noticed this line in the docs (generated from the help message)
```
 -v, --verbosity Reduce verbosity of error output.
```
because usually it works the other way. After a brief check, I see that indeed MORE information is displayed when `-v` option is provided.

### Checklist
- [x] All tests passing
- [x] Added a changelog entry

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 `Unreleased`_ - TBD
 -------------------
 
+**Fixed**
+
+- Description of ``-v`` or ``--verbosity`` option for CLI.
+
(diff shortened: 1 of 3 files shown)


Stranger6667:
Hi @lina-is-here !

Nice catch! Thanks :)

lina-is-here:
@Stranger6667 after another look, I think it's good to remove "error" from the message, because as I see in the code, it increases information not only about the errors, e.g. platform

Stranger6667:
Yep! You're right!

PR #1626, opened 2022-10-18
fix: Appending an extra slash to the `/` path

### LINKED ISSUE #1625: [BUG] Schemathesis generates tests for path `//` when testing path `/`.
## Checklist

- [x] I checked the [FAQ section](https://schemathesis.readthedocs.io/en/stable/faq.html#frequently-asked-questions) of the documentation
- [x] I looked for similar issues in the [issue tracker](https://github.com/schemathesis/schemathesis/issues)

## Describe the bug

Schemathesis generates the URL path `//` for the path `/`, which naturally triggers a 404 error, which causes a test failure.

My server does not "normalize" URL paths and I don't intend to support normalized URL paths, so 404 is expected behavior. Moreover, I don't define a 404 response on my `/` path because the `/` path itself can never return 404, so it wouldn't make sense to define a 404 response for it!

## To Reproduce

Create a trivial FastAPI application:

```python
from fastapi import FastAPI, Request, Response


app = FastAPI(
    title='Silly',
    version='1.0.0',
)


@app.get('/', response_class=Response)
def empty() -> Response:
    return Response()
```

Set up the Python environment::

```bash
python -m venv example-venv
./example-venv/bin/python -m pip install -U setuptools wheel
./example-venv/bin/pip install -U fastapi uvicorn schemathesis
```

Run Schemathesis:

```python
PYTHONPATH=. ./example-venv/bin/schemathesis run \
  --checks all --app 'example:app' '/openapi.json'
```

The schema itself:

```json
{
    "info": {
        "title": "Silly",
        "version": "1.0.0"
    },
    "openapi": "3.0.2",
    "paths": {
        "/": {
            "get": {
                "operationId": "empty__get",
                "responses": {
                    "200": {
                        "description": "Successful Response"
                    }
                },
                "summary": "Empty"
            }
        }
    }
}
```

If I change `/` to `/hello`, the tests pass successfully, so this appears to be a URL construction bug.

## Expected behavior

I expected Schemathesis to generate tests only for `/`, not for `//`.

Alternatively, I expected a configuration option that would allow me to disable testing of un-normalized URL paths. I looked carefully over the docs and didn't find anything along these lines.

- OS: Mac
- Python version: CPython 10.3.7
- Schemathesis version: 3.14.2
- Spec version: OpenAPI 3.0.2

## Additional context

In the short term, is there a way to temporarily exclude the path `/` from the generated test suite? It's not a highly important endpoint in our app (I'm OK foregoing test coverage on it), but it's also not something I can remove entirely (some usages do depend on it).

Stranger6667:
Fixes #1625

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 `Unreleased`_ - TBD
 -------------------
 
+**Fixed**
+
+- Appending an extra slash to the ``/`` path. `#1625`_
+
 .. _v3.17.3:
 
 `3.17.3`_ - 2022-10-10
@@ -3106,6 +3110,7 @@ Deprecated
 .. _0.3.0: https://github.com/schemathesis/schemathesis/compare/v0.2.0...v0.3.0
 .. _0.2.0: https://github.com/schemathesis/schemathesis/compare/v0.1.0...v0.2.0
 
+.. _#1625: https://github.com/schemathesis/schemathesis/issues/1625
 .. _#1614: https://github.com/schemathesis/schemathesis/issues/1614
 .. _#1592: https://github.com/schemathesis/schemathesis/issues/1592
 .. _#1591: https://github.com/schemathesis/schemathesis/issues/1591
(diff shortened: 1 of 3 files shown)


Stranger6667:
I was thinking about using `yarl` but didn't have the capacity to look into the issue as there were not so many issues with the URL construction logic and it is a pretty small surface (there were some complains about having too many dependencies). Not sure at the moment

PR #1675, opened 2023-02-01
fix: Replace `starlette.testclient.TestClient` with `starlette_testclient.TestClient` to keep compatibility with newer `starlette` versions

Stranger6667:
cc @Kludex 

As you suggested :) Works like a charm, thank you!


Kludex:
If you ever need something, let me know. :pray:

PR #1737, opened 2023-05-08
feat: add excluded_checks option

grogersxyz:
### Description

This feature allows users to pass a tuple of check functions that they want to be excluded. Exclusions are applied as a final step so adding an additional_check and excluding it will result in the check not being run.

### Checklist

- [x] All tests passing
- [x] Added a changelog entry
- [x] Extended the README / documentation, if necessary

### DIFF
--- a/src/schemathesis/models.py
+++ b/src/schemathesis/models.py
@@ -447,6 +447,7 @@ class Case:
         response: GenericResponse,
         checks: Tuple["CheckFunction", ...] = (),
         additional_checks: Tuple["CheckFunction", ...] = (),
+        excluded_checks: Tuple["CheckFunction", ...] = (),
         code_sample_style: Optional[str] = None,
     ) -> None:
         """Validate application response.
@@ -463,6 +464,8 @@ class Case:
         from .checks import ALL_CHECKS
 
         checks = checks or ALL_CHECKS
+        checks = tuple(check for check in checks if check not in excluded_checks)
+        additional_checks = tuple(check for check in additional_checks if check not in excluded_checks)
         failed_checks = []
         for check in chain(checks, additional_checks):
             copied_case = self.partial_deepcopy()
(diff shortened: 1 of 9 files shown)


src/schemathesis/models.py
```diff
@@ -463,6 +464,8 @@ def validate_response(
         from .checks import ALL_CHECKS
 
         checks = checks or ALL_CHECKS
+        checks = tuple(set(checks) - set(excluded_checks))
```
Stranger6667, line 467: I'd suggest using something like this to maintain a stable checks order which appears later on in failure reporting:

```python
checks = tuple(check for check in checks if check not in excluded_checks)
```

P.S. I assume it should be fine not to use sets as the container size is expected to be just a few elements

grogersxyz: Awesome suggestion, that makes it much easier to read as well, thanks! I've made those changes and amended that commit to include these changes (also made the same change in schemas.py:73

Stranger6667:
Thank you! A great change! :)

Would you be up to add a similar feature to Schemathesis CLI too?

grogersxyz:
I wasn't sure where the best place was to put the tests for the `exclude_checks` argument. I've added a test that checks the tests are excluded to the `cli/test_commands.py` and a test to make sure that the checks aren't run in cli/test_checks.py` let me know if you think those aren't in the right place and I can move them around :)

grogersxyz:
Apologies, I pushed `87f21e3` from a computer that didn't have pre-commit installed. Have just amended the commit to fix that

Stranger6667:
no worries! Thank you so much for implementing this feature! I am a bit concerned about adding the short `-e` for this as I was thinking about some CLI redesign in the near future. We could add it in the end though, after the redesign is done. What do you think about removing the short `-e` for the time being?

Stranger6667 (approved):

PR #1773, opened 2023-09-26
chore: update `pre-commit`

Stranger6667:


pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     "httpx>=0.22.0,<1.0",
     "hypothesis>=6.31.6,<7",
     "hypothesis_graphql>=0.9.0,<1",
-    "hypothesis_jsonschema>=0.22.1,<1",
+    "hypothesis_jsonschema>=0.22.1,<0.23",
```
allanlewis, line 41: **question:** What's the reason for this change? I use `hypothesis-jsonschema` explicitly in my code so I've had to restrict it to `<0.23` for the time being.

Stranger6667: I guess that I expected that the changes in `hypothesis-jsonschema` are not compatible with Schemathesis for now. Need to check it

allanlewis: OK, that makes sense. It's just a bit confusing given that previously this was much looser, and I didn't see anything in the changelog.

Perhaps it's worth pushing the maintainers of `hypothesis-jsonschema` to release a 1.0 and follow SemVer from there 🤔

Stranger6667: We plan to overhaul the internals there soon (to finally solve the recursive reference issue), so I hope it will happen in the foreseeable future

allanlewis: Great, thanks!

PR #1824, opened 2023-10-12
docs: add why is localhost slow on windows to faq

### LINKED ISSUE #1691: [BUG] Schemathesis extremly slow with localhost on windows
**Checklist**

- [X] I checked the [FAQ section](https://schemathesis.readthedocs.io/en/stable/faq.html#frequently-asked-questions) of the documentation
- [X] I looked for similar issues in the [issue tracker](https://github.com/schemathesis/schemathesis/issues)

**Describe the bug**
When using localhost to acces the OpenAPI schema like that `st run  http://localhost:5000/openapi.json` it take about one second for every requests. 

I think this is caused by the library **urllib3** as described in this [stackoverflow  question](https://stackoverflow.com/questions/47766158/why-is-python-requests-to-localhost-slow) : 

> the slow connections on Windows were related to urllib3 (the transport library used by requests) attempting an IPv6 connection first, which is refused since the server is listening on IPv4 only. For Windows only, however, there is an unavoidable 1 sec timeout on socket.connect for refused TCP connections, while the OS 'helpfully' retries the connection up to 3 times. (See: [Why do failed attempts of Socket.connect take 1 sec on Windows?](https://stackoverflow.com/questions/19440364/why-do-failed-attempts-of-socket-connect-take-1-sec-on-windows)).
> 
> On Linux, socket.connect fails immediately if the connection is refused, and it can attempt the IPv4 connection without any delay. In addition, some software, like curl, support limiting name resolution to only IPv4 or IPv6. Unfortunately, requests and urllib3 do not support this, and do not seem to plan to support this anytime soon (See: https://github.com/psf/requests/issues/1691)

**To Reproduce**

1. Run this command  `st run  http://localhost:5000/openapi.json` on **Windows**
2. Look at only one request being received by your api every second. 

**Expected behavior**
Pointing to localhost should give the save result as pointing to 127.0.0.1 or else should be documented to avoid any confusion for new users.

- OS: Windows
- Python version: 3.10.4 and 3.11
- Schemathesis version: 3.18.3
- Spec version: Open API 3.0.3

TheBigRoomXXL:
### Description

Update documentation in accordance with issue #1691 

Closes #1691

### Checklist

- [X] All tests passing
- [ ] Added a changelog entry
- [X] Extended the README / documentation, if necessary

Precedent `docs` commits were not added to the changelog so I gessed I should not do it either. I am correct?  

### Feedback on contributing documentation

This is not directly related to the content of the PR but I wanted to share some issues I had while contributing documentation so that it can be fixed. This my first time using Sphinx so maybe it's just something I didn't do properly.

1 - There is no instruction on how  to preview the changes to the documentation in the contribution guidelines. Adding a simple section with something like `cd docs/ && make html && python -m http.server -d _build/html/` would make it easier.  

2 - `sphinx-click` and `sphinx-rtd-theme` are not in the pyproject.toml docs dependencies and I couldn't build the documentation without installing them manually. Maybe they should be added to the dependency list?

Thank you for this awesome project! It has help me catch a lot of errors and I love using it!

### DIFF
--- a/docs/continuous_integration.rst
+++ b/docs/continuous_integration.rst
@@ -26,8 +26,8 @@ GitHub Actions
       # Runs Schemathesis tests with all checks enabled
       - uses: schemathesis/action@v1
         with:
-          # Your API schema location
-          schema: 'http://localhost:5000/api/openapi.json'
+          # Your API schema locationW
+          schema: 'http://127.0.0.1:5000/api/openapi.json'
           # OPTIONAL. Your Schemathesis.io token
           token: ${{ secrets.SCHEMATHESIS_TOKEN }}
 
@@ -70,7 +70,7 @@ GitLab CI
 
     variables:
       # API Schema location
-      API_SCHEMA: 'http://localhost:5000/api/openapi.json'
+      API_SCHEMA: 'http://127.0.0.1:5000/api/openapi.json'
       # OPTIONAL. Your Schemathesis.io token
       SCHEMATHESIS_TOKEN: ${{ secrets.SCHEMATHESIS_TOKEN }}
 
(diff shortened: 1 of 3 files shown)


docs/continuous_integration.rst
```diff
@@ -26,8 +26,8 @@ GitHub Actions
       # Runs Schemathesis tests with all checks enabled
       - uses: schemathesis/action@v1
         with:
-          # Your API schema location
-          schema: 'http://localhost:5000/api/openapi.json'
+          # Your API schema locationW
```
Stranger6667, line 29: Looks like a typo - `W` at the end of line :)

Stranger6667:
Thank you so much for contributing this PR!

> Precedent docs commits were not added to the changelog so I gessed I should not do it either. I am correct?

You are right! we may do it in the future, though, I am not sure yet.

I think that these are great points on the contributing documentation, I agree with all of them - will address them soon separately.

> Thank you for this awesome project! It has help me catch a lot of errors and I love using it!

Glad to hear it! :) Should you have any additional feedback or ideas, don't hesitate to create an issue and tag me :)

P.S. I'll adjust the typo in a separate commit so we can proceed now.

PR #1974, opened 2024-01-22
fix: validate that the graphql response is valid

devkral:
### Description

Currently call_and_validate doesn't check if in the graphql response are any errors or the answer is json at all.
This happened in my tests in my project.

This PR should fix the problem but I cannot test it.

It is really hard to contribute.
I don't understand in which test file I can define such a test and the test procedure fails for different sized terminals.

Sorry.

### Checklist

- [ ] Added changelog entry (follow guidelines in CONTRIBUTING.rst)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,7 @@ Changelog
 
 - ``--hypothesis-no-phases`` CLI option to disable Hypothesis testing phases. :issue:`1324`
 - Support for loading GraphQL schemas from JSON files that contain the ``__schema`` key.
+- Basic GraphQL response checking that only covers whether the response payload is valid JSON.
 
 .. _v3.24.3:
 
--- a/src/schemathesis/checks.py
+++ b/src/schemathesis/checks.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 from typing import TYPE_CHECKING
+import json
 
 from . import failures
-from .exceptions import get_server_error
+from .exceptions import get_server_error, get_response_parsing_error
 from .specs.openapi.checks import (
     content_type_conformance,
     response_headers_conformance,
@@ -17,10 +18,20 @@ if TYPE_CHECKING:
 
 def not_a_server_error(response: GenericResponse, case: Case) -> bool | None:
     """A check to verify that the response is not a server-side error."""
+    from .specs.graphql.schemas import GraphQLCase
+    from .transports.responses import get_json
+
     status_code = response.status_code
     if status_code >= 500:
         exc_class = get_server_error(status_code)
         raise exc_class(failures.ServerError.title, context=failures.ServerError(status_code=status_code))
+    if isinstance(case, GraphQLCase):
+        try:
+            _ = get_json(response)
+        except json.JSONDecodeError as exc:
+            exc_class = get_response_parsing_error(exc)
+            context = failures.JSONDecodeErrorContext.from_exception(exc)
+            raise exc_class(context.title, context=context) from exc
     return None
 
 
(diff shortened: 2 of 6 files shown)


Stranger6667:
First, I need to say that I am sorry about the suboptimal contribution experience. There are certain things there that I postponed fixing because they work without tox and some flaky tests are just cumbersome to fix properly. The test structure is also quite outdated.

#1975 sort of solves the issue with terminal width, so it should be easier now with running tests.

Re: PR. I think it is a great start!

### Response parsing

As we support multiple transports in that context, e.g. httpx and WSGI responses, then the first part might look like this (based on [this](https://github.com/schemathesis/schemathesis/blob/master/src/schemathesis/specs/openapi/schemas.py#L570C1-L579C73)):

```python
try:
    if isinstance(response, (requests.Response, httpx.Response)):
        data = json.loads(response.text)
    else:
        data = response.json
except JSONDecodeError as exc:
    exc_class = get_response_parsing_error(exc)
    context = failures.JSONDecodeErrorContext.from_exception(exc)
    raise exc_class(context.title, context=context) from exc
```

### Validation

I am not aware if the `errors` field is standardized, so I'd go maybe with a slightly more defensive approach:

```python
if data.get("errors", NOT_SET) is not NOT_SET:
    exc_class = get_exception("GraphQLErrors")
    raise exc_class(
        failures.GraphQLErrors.title,
        context=failures.GraphQLErrors(
            message="???",
        ),
    )
```

If there is some standard and way of structuring them, then we can distinguish them by serializing them into a canonical representation (here I mean that it does not depend on different order of errors, etc) and creating a separate exception class.

Also the sample above uses `failures.GraphQLErrors`, which is just a holder for additional context about the error so it can be formatted later on. For now, it could be something generic, but if there is some common way to structure GraphQL errors, then we can use it and display them nicely in CLI

Also, this function should not fail if the case is not a `GraphQLCase` instance, just return `True` in such a case.

This function should appear in `DEFAULT_CHECKS` inside `src/checks.py`

### Tests

I think I am going to write a few of them myself + clean up a few things, so you don't have to go through this bumpy road. Then, as a separate step, I'll update the contribution guide so it is more clear on what tests should go where.

Thank you @devkral for bringing this into Schemathesis and your persistence in the issue tracker. I genuinely appreciate it :)

Stranger6667:
Also, A bit more context. Initially, I wanted to add something like this, but it wasn't clear to me how to distinguish client (i.e. syntactically invalid queries) & server errors. My understanding is that every framework does something custom, hence I stopped researching and left the absolutely minimal version that just checks for the 5xx status code.

I want to avoid this check reporting client-side errors as test failures, so the user can just focus on server errors. Client-side errors its something Schemathesis should fix (or the user if it is coming from an extension), so they could be reported as hard errors in Schemathesis implementation.

Stranger6667:
@devkral 

What do you think about it? If you can point me to some common practices on how the "errors" value is structured, it would help me a lot

Stranger6667:
Also, we can split the "errors" part out and just add a check that parses JSON and extend it later on. It might be simpler and will also reduce the total number of code to review

devkral:
the errors array normally doesn't appear, but to be sure to handle an empty array of an unknown application

devkral:
I wouldn't make more assumptions because some server errors are wanted (e.g. PermissionDenied) but I didn't figure it out either how to seperate client and server errors.

Stranger6667:
Ok, let me add a few things to this PR and merge it, the errors part could be done separately. Meanwhile I'll research more on the error structure

devkral:
Thank you!

Stranger6667:
Btw, I have a draft for it and hopefully it will be finished in a few days

PR #1985, opened 2024-01-23
feat: Shortcut to create strategies for all operations or a subset of them

### LINKED ISSUE #1982: [FEATURE] allow selecting Type as strategy for graphql schemas
Thanks for the great improvements to the graphql schemas. I have a little suggestion how to select the operation:

Would it be possible to subclass the dict used for graphql_schema["Operation"] and provide a as_strategy method?

This would allow intuitively selecting an operation and test all fields.

Example:

```python
graphql_schema["Query"].as_strategy()

```
This shortcut would maybe be nice for openapi too and could allow feature parity with the existing @parameterize decorator.

Further thought: would it also be possible to implement an as_strategy method directly on a schema (to test everything without selecting)?

Stranger6667:
Resolves #1982

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 :version:`Unreleased <v3.24.1...HEAD>` - TBD
 --------------------------------------------
 
+**Added**
+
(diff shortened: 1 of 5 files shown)


devkral:
thanks

PR #2071, opened 2024-02-26
Use the new-style pytest hook wrappers.

kbaikov:
This avoids the pytest:
``` PluggyTeardownRaisedWarning: A plugin raised an exception during an old-style hookwrapper teardown. Plugin: schemathesis, Hook: pytest_pyfunc_call
```

🚨Please review the [Contributing Guidelines](https://github.com/schemathesis/schemathesis/blob/master/CONTRIBUTING.rst).

### Description

Use the new style pytest plugin hooks

### Checklist

- [ ] Added changelog entry (follow guidelines in CONTRIBUTING.rst)


dongfangtianyu:
Hi, @kbaikov 

To the best of my knowledge, pytest does not intend to deprecate the old-style wrappers, so we can continue using the old-style wrappers.

Moreover, the difference between the new and old styles is not only in the changes to parameters but also in changes to usage.

This Pull Request only modifies the parameter names without changing the usage, which seems to be an incorrect approach for the new-style wrapper.

PR #2134, opened 2024-04-14
new version pytest_subtests

Enimalojd:
Fixed some tests output.
Replaced the nullcontext.
I marked two tests with the skip label, which I couldn't fix in any way.


Stranger6667:
Resolved in #2498

PR #2279, opened 2024-06-22
Fix openapi relative ref

frederikaalund:
### Description

This fixes Internal error during $ref resolve when using relative file paths.

### Checklist

- [x] Added changelog entry (follow guidelines in CONTRIBUTING.rst)


Stranger6667:
Hi! Thank you, could you add a test case for this fix? Or share the schema you have (anonymised will work)

frederikaalund:
I added a test that fails without this PR. I included one of the paths (`user_account.yaml`) in my schema along with all it's $refs (`parameters.yaml`, `schemas.yaml`, `responses.yaml`). I can simplify the example further if you'd like but for now I kept all details to make it more "real".

Stranger6667:
Thank you so much! Yes, lets keep it as is :)

Stranger6667 (approved):

PR #2936, opened 2025-07-07
Add encoding to junit output

kayoub5:
🚨Please review the [Contributing Guidelines](https://github.com/schemathesis/schemathesis/blob/master/CONTRIBUTING.md).

### Description

Briefly describe your changes.

### Checklist

- [x] Added changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased](https://github.com/schemathesis/schemathesis/compare/v4.0.5...HEAD) - TBD
 
+### :bug: Fixed
+- Use `utf-8` encoding when generating JUnit reports
+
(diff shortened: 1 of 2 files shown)


Stranger6667:
In what case does the absence of `encoding` cause problems? It would be nice to have a test case

kayoub5:
> In what case does the absence of `encoding` cause problems? It would be nice to have a test case

Windows, schemathesis tried to print some special characters to junit report, 

The `cp1252` was being used by default.

Default python encoding in windows can be found with
```
import locale
print(locale.getpreferredencoding(False))
```

Stranger6667:
I've seen issues when IO encoding was set to `cp1252` by default, and there is already some code to handle stdout / yaml serialization in such cases. However, I haven't thought about JUnit, but it should essentially be the same problem. This should go away in Python 3.15, hopefully (at least the majority of issues with the default encoding), but I am happy to merge these changes as is right now. Thank you for bringing this up!

PR #3033, opened 2025-08-13
feat: Open API links inference from Location headers

### LINKED ISSUE #2953: [FEATURE] Basic inference of Open API links in runtime
The task here is to add some infrastructure to allow for adding new Open API links after the testing process has started. Then add links based on the `Location` header in `POST` responses.

Stranger6667:
Resolves #2953

- [x] Run on 201 responses on POST API operations during `examples` / `coverage` / `fuzzing` phases, unless there is already a link
- [ ] If other phases are disabled - run a quick set of API calls on POST / PUT (with filters applied) to get `Location` values (or postpone it to another PR?)
- [x] For errors in inferred links, the messaging should be different - it is a bug in Schemathesis if they don't work properly
- [x] Ensure inference takes the filters into account.
- [x] Investigate test flakiness
- [x] Base path can be present in relative urls too
- [ ] Check the logic manually on DRF
- [x] Display number of inferred links in CLI (use the `x-inferred` attribute)
- [x] Maybe just have “algorithms” config instead of “enable”? To avoid having two ways to disable it
- [x] Document this feature (mention popular frameworks that add `Location` automatically)
- [x] Consider adding links to other API operations (e.g. `PATCH` / `PUT` with the same prefix)
- [x] Config option to enable / disable it for the stateful phase? Useful at least for testing
- [x] Improve wording in the docs
- [x] Strip base path from `Location`
- [x] If `Location` contains query parameters - add it to the generated link?
- [x] Test - there should be a bug in GET, which should be discovered via an automatically inferred link
- [x] Test - parameters extraction in isolation (evaluate that runtime expression)


src/schemathesis/engine/context.py
```diff
@@ -65,6 +79,21 @@ def is_interrupted(self) -> bool:
     def has_reached_the_failure_limit(self) -> bool:
         return self.control.has_reached_the_failure_limit
 
+    def update_phase(self, phase: Phase) -> None:
```
Stranger6667: Not sure if the best name + missing some comments / docs

src/schemathesis/engine/context.py
```diff
@@ -65,6 +79,21 @@ def is_interrupted(self) -> bool:
...
+            from schemathesis.specs.openapi.stateful import inference
+
+            assert isinstance(self.schema, BaseOpenAPISchema)
+
+            inferencer = inference.LinkInferencer.from_schema(self.schema)
+            injected = 0
+            for operation, entries in self.repository.location_headers.items():
+                injected += inferencer.inject_links(operation.definition.raw, entries)
+            if injected:
+                phase.is_enabled = True
```
Stranger6667: Move to a separate method?

src/schemathesis/engine/phases/unit/__init__.py
```diff
@@ -76,6 +76,8 @@ def execute(engine: EngineContext, phase: Phase) -> events.EventGenerator:
                                 status = event.status
                             if event.status in (Status.ERROR, Status.FAILURE):
                                 engine.control.count_failure()
+                            if engine.repository.is_enabled:
```
Stranger6667: Move to a separate property with better name

src/schemathesis/engine/phases/unit/__init__.py
```diff
@@ -76,6 +76,8 @@ def execute(engine: EngineContext, phase: Phase) -> events.EventGenerator:
...
+                                engine.repository.process_recorder(event.recorder)
```
Stranger6667: Name is meh

src/schemathesis/engine/core.py
```diff
@@ -122,6 +124,8 @@ def execute(self, engine: EngineContext) -> EventGenerator:
             for phase in self.phases:
                 if engine.has_reached_the_failure_limit:
                     phase.skip_reason = PhaseSkipReason.FAILURE_LIMIT_REACHED
+                # Phase can be enabled if certain conditions are met
+                engine.update_phase(phase)
```
Stranger6667: Improve naming & comment

src/schemathesis/engine/core.py
```diff
@@ -26,8 +27,9 @@ def execute(self) -> EventStream:
         if self.schema.config.auth.is_defined:
             unregister_auth()
 
-        ctx = EngineContext(schema=self.schema, stop_event=threading.Event())
         plan = self._create_execution_plan()
+        repository = DataRepository(list(plan.phases))
+        ctx = EngineContext(schema=self.schema, repository=repository, stop_event=threading.Event())
```
Stranger6667: Not sure if it is good to pass only phases there - maybe it will be better to make a decision about recording data not in the repository itself, but in EngineContext. E.g. repository is always the same,  but the context decides whether to write to the repository and what to write there

src/schemathesis/engine/repository.py
```diff
@@ -0,0 +1,37 @@
...
+from schemathesis.engine.phases import Phase, PhaseName, PhaseSkipReason
+from schemathesis.engine.recorder import ScenarioRecorder
+from schemathesis.schemas import APIOperation
+
+
+@dataclass
+class DataRepository:
+    """Repository for data extracted from API responses to improve test generation."""
+
+    location_headers: dict[APIOperation, list[LocationHeaderEntry]]
```
Stranger6667: Is it possible to avoid this dependency somehow? maybe store id or label if it is not too slow to lookup

src/schemathesis/specs/openapi/stateful/inference.py
```diff
@@ -0,0 +1,247 @@
...
+    def to_link_base(self) -> dict[str, Any]:
+        return {"operationRef": self.value, "x-inferred": True}
+
+
+Endpoint = Union[EndpointById, EndpointByRef]
+# Method, path, response code, sorted path parameter names
+SeenLinkKey = tuple[str, str, int, tuple[str, ...]]
+
+
+@dataclass
```
Stranger6667, line 61: Docstrings for this and other classes

src/schemathesis/specs/openapi/stateful/inference.py
```diff
@@ -0,0 +1,247 @@
...
+    def match(self, path: str) -> tuple[Endpoint, Mapping[str, str]] | None:
+        """Match path to endpoint and extract path parameters."""
+        try:
+            return self._adapter.match(path)
+        except (NotFound, MethodNotAllowed):
+            return None
+
+    def build_links(self, location: str) -> list[dict]:
+        """Build all possible OpenAPI link definitions from Location header."""
+        # TODO: Move to tests?
```
Stranger6667: Move to tests indeed

src/schemathesis/specs/openapi/stateful/inference.py
```diff
@@ -0,0 +1,247 @@
...
+        injected = 0
+
+        for entry in entries:
+            location = self._normalize_location(entry.value)
+            if location is None:
+                continue
+
+            matches = self._find_matches_from_normalized_location(location)
+            if matches is None:
+                continue
```
Stranger6667: Add comments on when `continue` happens

src/schemathesis/specs/openapi/stateful/inference.py
```diff
@@ -0,0 +1,247 @@
...
+
+            key = (matches.exact.method, matches.exact.path, entry.status_code, tuple(sorted(matches.parameters)))
+            if key in seen:
+                continue
+            seen.add(key)
+            # Find the right bucket for the response status
```
Stranger6667: "Or create a new one"

test/engine/test_probes.py
```diff
@@ -8,11 +8,12 @@
 from schemathesis.core.transport import USER_AGENT
 from schemathesis.engine.context import EngineContext
 from schemathesis.engine.phases import probes
+from schemathesis.engine.repository import DataRepository
 
 
 @pytest.fixture()
 def ctx(openapi_30):
-    return EngineContext(schema=openapi_30, stop_event=threading.Event())
+    return EngineContext(schema=openapi_30, repository=DataRepository([]), stop_event=threading.Event())
```
Stranger6667: It will be nice to avoid passing `DataRepository` there.

src/schemathesis/engine/repository.py
```diff
@@ -0,0 +1,37 @@
+from dataclasses import dataclass
+
+from schemathesis.core.repository import LocationHeaderEntry
+from schemathesis.engine.phases import Phase, PhaseName, PhaseSkipReason
+from schemathesis.engine.recorder import ScenarioRecorder
+from schemathesis.schemas import APIOperation
+
+
+@dataclass
+class DataRepository:
```
Stranger6667: Maybe a bit clearer naming which will give a clearer picture of what it does

PR #3101, opened 2025-09-25
refactor: Resolve response schemas eagerly

Stranger6667:


src/schemathesis/schemas.py
```diff
@@ -628,6 +628,7 @@ class APIOperation(Generic[P]):
     cookies: ParameterSet[P] = field(default_factory=ParameterSet)
     query: ParameterSet[P] = field(default_factory=ParameterSet)
     body: PayloadAlternatives[P] = field(default_factory=PayloadAlternatives)
+    responses: dict[str, Any] = field(default_factory=dict)
```
Stranger6667: Set proper types - maybe some protocol + generic?

src/schemathesis/core/jsonschema/bundler.py
```diff
@@ -135,7 +136,8 @@ def bundle_recursive(current: JsonSchema | list[JsonSchema]) -> JsonSchema | lis
 
         assert isinstance(bundled, dict)
 
-        if "$ref" in bundled and len(defs) == 1:
+        # Inlining such a schema is only possible if recursive references were inlined
+        if inline_recursive and "$ref" in bundled and len(defs) == 1:
```
Stranger6667: It would be nice to still do this inlining step for cases when "inline_recursive=False`, but there were no recursive references. It will make validation of responses a bit faster. Add a test for it

src/schemathesis/specs/openapi/_adapter.py
```diff
@@ -25,9 +27,54 @@ def prepare_parameters(item: PathItem | Operation, *, resolver: RefResolver, bun
...
         yield definition
+
+
+@dataclass
+class Response:
+    """OpenAPI response definition."""
+
+    status_code: str
+    definition: dict[str, Any]
+    # TODO: Use types from `core.jsonschema`
```
Stranger6667: Indeed

src/schemathesis/specs/openapi/_adapter.py
```diff
@@ -25,9 +27,54 @@ def prepare_parameters(item: PathItem | Operation, *, resolver: RefResolver, bun
...
+    schema: dict[str, Any] | bool | None
```
Stranger6667: Store `scope` and bundle it lazily

src/schemathesis/specs/openapi/_adapter.py
```diff
@@ -25,9 +27,54 @@ def prepare_parameters(item: PathItem | Operation, *, resolver: RefResolver, bun
...
+
+    __slots__ = ("status_code", "definition", "schema")
```
Stranger6667: Also store `headers`

src/schemathesis/specs/openapi/_adapter.py
```diff
@@ -25,9 +27,54 @@ def prepare_parameters(item: PathItem | Operation, *, resolver: RefResolver, bun
...
+    _inner: dict[str, Response]
+
+    __slots__ = ("_inner",)
+
+    def find_by_status_code(self, status_code: int) -> Response | None:
+        """Find the most specific response definition matching the given HTTP status code."""
+        return _find_by_status_code(self._inner, status_code)
+
+
+T = TypeVar("T")
```
Stranger6667: It may be unnecessary to have this generic in the future - the function below is needed for inference and works with `dict`, but it actually can work with `Response` and inject links there. Add a note for the future

src/schemathesis/specs/openapi/converter.py
```diff
@@ -108,9 +108,23 @@ def is_read_only(schema: dict[str, Any] | bool) -> bool:
...
+) -> dict[str, Any]: ...  # pragma: no cover
+
+
+@overload
+def to_json_schema_recursive(
+    schema: bool, nullable_name: str, is_response_schema: bool = False, update_quantifiers: bool = True
+) -> bool: ...  # pragma: no cover
+
+
+def to_json_schema_recursive(
```
Stranger6667, line 130: Move to `core.jsonschema` & refactor? maybe move `update_quantifier` to `core` too?

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -414,6 +417,9 @@ def make_operation(
         self.dispatch_hook("before_init_operation", HookContext(operation=operation), operation)
         return operation
 
+    def _parse_responses(self, definition: dict[str, Any], scope: str) -> Responses:
```
Stranger6667: It will be nice to pass a spec-specific `adapter` and offload all implementation to it instead of doing inheritance stuff.

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -502,27 +504,14 @@ def get_parameter_serializer(self, operation: APIOperation, location: str) -> Ca
...
         self, operation: APIOperation, response: Response
     ) -> tuple[list[str], dict[str, dict[str, Any]] | None] | None:
-        resolved = self._get_response_definitions(operation, response)
-        if not resolved:
+        definition = operation.responses.find_by_status_code(response.status_code)
+        if definition is None:
             return None
-        scopes, definitions = resolved
-        return scopes, definitions.get("headers")
+        # TODO: It should be proper scopes / resolve it eagerly
```
Stranger6667, line 520: Indeed

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -797,16 +762,20 @@ def get_strategies_from_examples(self, operation: APIOperation, **kwargs: Any) -
...
-        scopes, definition = resolve_in_scope(self.resolver, definition, scope)
-        schema = definition.get("schema")
-        if not schema:
-            return scopes, None
-        # Extra conversion to JSON Schema is needed here if there was one $ref in the input
-        # because it is not converted
-        return scopes, to_json_schema_recursive(
-            schema, self.nullable_name, is_response_schema=True, update_quantifiers=False
-        )
+    def _parse_responses(self, definition: dict[str, Any], scope: str) -> Responses:
```
Stranger6667: Move this into "adapter"

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -974,29 +943,37 @@ def collect_parameters(
...
-        option = next(options, None)
-        # "schema" is an optional key in the `MediaType` object
-        if option and "schema" in option:
-            # Extra conversion to JSON Schema is needed here if there was one $ref in the input
-            # because it is not converted
-            return scopes, to_json_schema_recursive(
-                option["schema"], self.nullable_name, is_response_schema=True, update_quantifiers=False
-            )
-        return scopes, None
+    def _parse_responses(self, definition: dict[str, Any], scope: str) -> Responses:
```
Stranger6667: Move this into "adapter"

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -974,29 +943,37 @@ def collect_parameters(
...
+        responses = {}
+        for key, response in definition.get("responses", {}).items():
+            scopes, response = resolve_in_scope(self.resolver, response, scope)
```
Stranger6667: re-check if something differs for Open API 3.1

src/schemathesis/specs/openapi/schemas.py
Stranger6667: Use a proper `types.v3` definition + in another method too

src/schemathesis/specs/openapi/schemas.py
```diff
@@ -974,29 +943,37 @@ def collect_parameters(
...
+            options = iter(response.get("content", {}).values())
```
Stranger6667: There could be more media types ... is it an oversight that the first one is taken? if so, note it and fix separately

Stranger6667 (commented):
- [ ] postpone resolving & transforming - add caching inside `Response` + store scope for it 
- [ ] Test for boolean response schema
- [ ] Resolve & store headers the same way and use them in the corresponding check

PR #3198, opened 2025-10-22
Chore: use st.one_of instead of home made combine

bruchar1:
As there are no reasons to reinvent the wheel.


Stranger6667:
Thanks! I somehow thought that `one_of` would not work for my original use case and never re-visited it

PR #3200, opened 2025-10-23
fix: do not require tomli for projects with python 3.11+

### LINKED ISSUE #3199: [FEATURE] Tomli package shouldn't be required for projects with python 3.11+
### Is your feature request related to a problem? Please describe

`tomllib` built-in package [was introduced](https://docs.python.org/3/library/tomllib.html) with Python 3.11.

I have a project that requires `python>=3.11` and includes `schemathesis` as a dependency. I would prefer not to include `tomli` as an additional dependency if it’s not necessary.

### Describe the solution you'd like

Add a conditional dependency such as `"tomli>=2.2.1; python_version < '3.11'"`, and use conditional imports or if-else statements in the parts of the code where `tomli` is currently used.

danfimov:
### Description

Added condition for including `tomli` to dependencies (do not require it for projects with python 3.11+)

Closes #3199 

### Checklist

- [x] Added changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased](https://github.com/schemathesis/schemathesis/compare/v4.3.10...HEAD) - TBD
 
+### :bug: Fixed
+
+- Tomli dependency is optional for Python 3.11+. [#3199](https://github.com/schemathesis/schemathesis/issues/3199)
+
 ## [4.3.10](https://github.com/schemathesis/schemathesis/compare/v4.3.9...v4.3.10) - 2025-10-22
 
 ### :bug: Fixed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Internal error when generating invalid type in the coverage phase.
 - Internal error when generating valid instance against an array of types in the coverage phase. [#3194](https://github.com/schemathesis/schemathesis/issues/3194)
 - Missing required request body for schemas without validation keywords.
+- Tomli dependency is optional for Python 3.11+. [#3199](https://github.com/schemathesis/schemathesis/issues/3199)
```
Stranger6667: It belongs to the `Unreleased` block

src/schemathesis/cli/commands/__init__.py
```diff
@@ -16,6 +17,11 @@
 from schemathesis.core.errors import HookError, format_exception
 from schemathesis.core.version import SCHEMATHESIS_VERSION
 
+if sys.version_info < (3, 11):
+    from tomli import TOMLDecodeError
```
Stranger6667, line 20: The original import is still there. See line 8 above, and it will fail on Python >=3.11 as `tomli` won't be installed

danfimov: Oh, missed it. Fixed)

CHANGELOG.md
danfimov: Fixed - moved to the desired block

Stranger6667 (approved):

Stranger6667:
Thank you! I plan to push some more changes before making a release. The next release is likely to land around this weekend.

PR #3202, opened 2025-10-23
feat: Get strategy from examples

bruchar1:
### Description

Adding a `as_examples()` method, similar to `as_strategy()`. The goal is to be able to easily test the examples from the schema in a pytest context. 

Tests and documentation are currently missing. I want to be sure first that the proposed function and naming are correct.

### Checklist

- [ ] Added changelog entry

### DIFF
--- a/src/schemathesis/schemas.py
+++ b/src/schemathesis/schemas.py
@@ -461,6 +461,26 @@ class BaseSchema(Mapping):
         ]
         return st.one_of(_strategies)
 
+    def as_examples(
+        self,
+        **kwargs: Any,
+    ) -> SearchStrategy:
+        """Create a list of test cases based on the examples defined in the schema for all operations.
+
+        Use with `@given` in non-Schemathesis tests.
+
+        Args:
+            **kwargs: Additional keywords for each strategy.
+
+        Returns:
+            Combined Hypothesis strategy for all valid operations in the schema.
+
+        """
+        _strategies = [
+            operation.ok().as_examples(**kwargs) for operation in self.get_all_operations() if isinstance(operation, Ok)
+        ]
+        return st.one_of(_strategies)
+
     def find_operation_by_label(self, label: str) -> APIOperation | None:
         raise NotImplementedError
 
@@ -501,6 +521,24 @@ class APIOperationMap(Mapping):
         _strategies = [
             operation.as_strategy(generation_mode=generation_mode, **kwargs) for operation in self._data.values()
         ]
+        return st.one_of(_strategies)
+
+    def as_examples(
+        self,
+        **kwargs: Any,
+    ) -> SearchStrategy:
+        """Create a list of test cases based on the examples defined in the schema for all operations in this subset.
+
+        Use with `@given` in non-Schemathesis tests.
+
+        Args:
+            **kwargs: Additional keywords for each strategy.
+
+        Returns:
+            Combined Hypothesis strategy for all valid operations in the schema.
+
+        """
+        _strategies = [operation.as_examples(**kwargs) for operation in self._data.values()]
         return st.one_of(_strategies)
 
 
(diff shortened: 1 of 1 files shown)


Stranger6667:
Right now examples from the schema are added as Hypothesis ‘examples’ decorator if the examples phase is enabled (in pytest), but you need more flexibility right? How would such a test look like?

src/schemathesis/schemas.py
```diff
@@ -461,6 +461,26 @@ def as_strategy(
         ]
         return st.one_of(_strategies)
 
+    def as_examples(
+        self,
+        **kwargs: Any,
+    ) -> SearchStrategy:
+        """Create a list of test cases based on the examples defined in the schema for all operations.
```
Stranger6667, line 468: As it returns a strategy, the details could be tweaked a bit. It is about a strategy that randomly takes from such a list of examples

bruchar1:
Well. What I want is probably a deterministic list of example `Case` more than a strategy.

If fact, what I'd really like would be a parametrized fixture generating all the example cases for an operation or for a whole schema. But I also want my schema to be lazy loaded in a fixture. Therefore, the only way I found to get what I want is to use a `@given(st.data())` along with a strategy. Probably I should rather loop through the cases and use subtests.

I will try to rewrite it based on `builder.generate_example_cases()`.

I realize that while I like the explorational behavior of schemathesis, I also need a deterministic test suite to run in the CI. For that part, I'm more interested by the fact that schemathesis can validate responses than on the randomness of the test cases.

Stranger6667:
Right now, the following would give you only schema examples in `schema.parametrize`:

```python
from hypothesis import settings, Phase

schema = ...  # load as usual

@schema.parametrize()
@settings(phases=[Phase.explicit])
def test_api(case):
    response = case.call()
    case.validate_response(response)
```

Add this to your `schemathesis.toml` to disable coverage/fuzzing/stateful phases:

```toml
[phases]
enabled = false           # Disable all phases by default
examples.enabled = true   # Enable only examples phase
```

**Note**: There will be minimal random generation for required parameters that lack examples in the spec (necessary to create valid test cases).

Would it be sufficient?

bruchar1:
The problem with `parametrize` is that it doesn't give me enough granularity. For example, I want to add markers to different parts of my API, and I need to handle some special cases with xfail, etc. 

What I'm doing right now:

```
def _get_api_operations() -> list[ParameterSet]:
    result = []
    for operation_result in schema.get_all_operations():
        if isinstance(operation_result, Ok):
            operation = operation_result.ok()
            if operation.tags:
                tag = operation.tags[0]
                marks = [getattr(pytest.mark, tag.lower())]
            else:
                tag = ""
                marks = []
            id_ = f"{tag} - {operation.method.upper()} {operation.path}"

            result.append(pytest.param(operation, id=id_, marks=marks))
    return result


@pytest.fixture(params=_get_api_operations(), scope="session")
def api_operation(request: pytest.FixtureRequest) -> APIOperation:
    return request.param


def _get_path_params(path: str) -> dict[str, str]:
    path_params = {}
    for match in re.finditer(r"\{([a-zA-Z_]+)}", path):
        path_param = match.group(1)
        path_params[path_param] = "0"
    return path_params


@settings(max_examples=1, deadline=None, suppress_health_check=[HealthCheck.nested_given])
@given(data=st.data())
def test_operation_example(api_operation, data):
    is_draft = api_operation.definition.raw.get("x-draft", False)

    examples = api_operation.get_strategies_from_examples()
    if not examples:
        if is_draft:
            pytest.xfail(reason=f"No examples for draft endpoint {api_operation.method} {api_operation.path}")

        pytest.fail(reason=f"No examples for {api_operation.method} {api_operation.path}")

    for example in examples:
        case = data.draw(example)
        case.headers.update(api_operation.schema.config.headers)
        if path_params := _get_path_params(api_operation.path):
            case.path_parameters.update(path_params)

        response = case.call(timeout=None)
        if response.status_code == 501 and is_draft:
            pytest.xfail(reason="Unimplemented endpoint.")

        api_operation.validate_response(response)
```

I think I will close that PR now, because I think my solution is good enough for now, and I'm still unsure what would be the easiest way to expose it from schemathesis public interface.

PR #3305, opened 2025-11-08
feat: Open API aware auth config

### LINKED ISSUE #1710: [FEATURE] Open API: named auth implementation
### Problem

It could be tedious to use different auth types within the same Open API schemas in Schemathesis. Essentially it means duplicating the filtering logic inside the auth implementation itself (filters will simplify it though).

### Solution

Add a method that makes Schemathesis use certain auth mechanisms based on their name & type.

Schema:

```yaml
components:
  securitySchemes:
    BasicAuth:
      type: http
      scheme: basic
    BearerAuth:
      type: http
      scheme: bearer
    ApiKeyAuth:
      type: apiKey
      in: header
      name: X-API-Key
    OpenID:
      type: openIdConnect
      openIdConnectUrl: https://example.com/.well-known/openid-configuration
    OAuth2:
      type: oauth2
      flows:
        password:
          refreshUrl: https://example.com/oauth/refresh
          tokenUrl: https://example.com/oauth/token
          scopes:
            read: Grants read access
            write: Grants write access
            admin: Grants access to admin operations
```

Then it could be something like:

```python
schema.auth.openapi.http_basic("BasicAuth", username="username", password="password")
schema.auth.openapi.http_bearer("BearerAuth", token="TOKEN")
schema.auth.openapi.api_key("ApiKeyAuth", api_key="API-KEY")
```

I.e. some specifically typed methods that will automatically add proper impl under the hood.

Alternative, a simpler interface, but typing would be harder to specify (+ some validation would be needed):

```python
schema.auth.openapi.define("BasicAuth", username="username", password="password")
schema.auth.openapi.define("BearerAuth", token="TOKEN")
schema.auth.openapi.define("ApiKeyAuth", api_key="API-KEY")
schema.auth.openapi.define("OAuth2", username="username", password="password")
```

Also, overriding some values would be nice for testing purposes - e.g. specify a different `refreshUrl` for testing purposes.

Probably Schemathesis can't support all security schemes, especially ones with device auth codes, but username / password or token-like should be a good starting point.

### Implementation notes

- I'd like to have everything called via `schema.auth`, but since `AuthStorage` is spec-agnostic we might need to subclass it in order to add some setup-time validation for inputs

Stranger6667:
Resolves #1710


docs/reference/warnings.md
```diff
@@ -92,7 +93,34 @@ When this warning appears, Schemathesis skips validation because it cannot deser
...
+- **Typo in scheme name**: Config has `ApiKeyHeadr` but schema defines `ApiKeyHeader` (Schemathesis suggests corrections)
+- **Scheme not defined**: Config references a scheme that doesn't exist in `components.securitySchemes`
+- **Scheme unused**: Scheme is defined but no operations reference it in their `security` array
+
+To fix:
+
+- Verify the scheme name in your config matches exactly what's in your OpenAPI schema's `securitySchemes`
+- Check that at least one operation uses the scheme in its `security` requirements
+- Remove unused auth configurations if they're no longer needed
+
```
Stranger6667: Could be way more concise

src/schemathesis/config/_auth.py
```diff
@@ -9,16 +9,109 @@
...
+    type: str
+    scheme: str
+    username: str
+    password: str
+
+    __slots__ = ("type", "scheme", "username", "password")
+
+    def __init__(self, *, type: str = "", scheme: str = "", username: str = "", password: str = "") -> None:
+        self.type = type
+        self.scheme = scheme
```
Stranger6667: Unused

PR #3452, opened 2026-01-11
chore: add UP ruff rule

even-even:
### Description

add https://docs.astral.sh/ruff/rules/#pyupgrade-up rule and fix in code

### Checklist

- [x] Added changelog entry
- [x] Updated README/documentation, if necessary

### DIFF
--- a/test/specs/openapi/stateful/inference/test_dependencies.py
+++ b/test/specs/openapi/stateful/inference/test_dependencies.py
@@ -1752,57 +1752,6 @@ def test_resource_name_from_path_with_param(path, param_name, expected):
     assert naming.from_path(path, param_name) == expected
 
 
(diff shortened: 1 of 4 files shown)


test/specs/openapi/stateful/inference/test_dependencies.py
```diff
@@ -1752,6 +1752,57 @@ def test_resource_name_from_path_with_param(path, param_name, expected):
     assert naming.from_path(path, param_name) == expected
 
 
+def test_remove_unused_resources():
```
Stranger6667: how is that relevant to the PR?

even-even: These tests were for debugging - I added them by accident.
Deleted it.

PR #3514, opened 2026-02-16
Support pyrate-limiter 4.x

Enimalojd:
- Update rate limiter construction to pyrate-limiter v4: use Limiter(rate, buffer_ms=50) instead of max_delay-based configuration.
- Remove _get_max_delay helper which was only used to derive max_delay in v3.
- Adapt tests to v4 behavior:
  - test_maximum_requests now asserts that 5 requests succeed under "5/h" without relying on BucketFullException or delay_or_raise.
  - Replace test_get_max_delay with test_parse_units to cover parse_units() on "1/s", "100/m" and "1000/h". References:
 - pyrate-limiter 4.0.0 changelog:
  - ExceptionsRemoved: BucketFullException and LimiterDelayException removed, rate limiting now uses blocking behavior instead of exception-based flow control.
  - Limiter constructor simplified: max_delay, raise_when_fail and clock parameters removed.
  - https://raw.githubusercontent.com/vutran1710/PyrateLimiter/v4.0.0/CHANGELOG.md


pyproject.toml
```diff
@@ -42,7 +42,7 @@ dependencies = [
   "jsonschema[format]>=4.18.0,<5.0",
   "jsonschema_rs>=0.41.0",
   "junit-xml>=1.9,<2.0",
-  "pyrate-limiter>=3.0,<4.0",
+  "pyrate-limiter>=4.0",
```
Stranger6667: I am hesitant on removing the upper bound - it happened already many times that new major release has backward incompat changes and users report it here. I'd like to avoid such UX and would prefer more conservative upper bound - e.g. `<5.0`

Stranger6667:
Thanks!

PR #3606, opened 2026-03-09
 feat: Adapt extracted-value reuse rate per stateful link

Stranger6667:


Stranger6667:
@codspeedbot fix this regression

PR #3621, opened 2026-03-19
Fix/coverage path param minlength

radionest:
Description

_ensure_valid_path_parameter_schema unconditionally set minLength to 1,
overwriting any stricter constraint from the original schema. This caused the
Coverage phase to generate path parameter values shorter than the schema
allows (e.g. "0" for a parameter with minLength: 5), leading to false positive
 positive_data_acceptance failures.

Fix: use max(schema.get("minLength", 0), 1) to preserve the original
constraint when it is already >= 1.

Checklist

- [x] Added changelog entry


Stranger6667 (approved):
Thanks!

PR #3622, opened 2026-03-19
Bugfix: apply_to filter_set not updated between hook registrations

elsurudo:
### Description

When multiple hooks are registered via `schemathesis.hook.apply_to()` with different filters, only the first hook receives the correct filter_set. All subsequent hooks silently inherit the first hook's filter, causing them to fire for the wrong operations.

#### Root Cause

In `to_filterable_hook` (src/schemathesis/hooks.py), the register closure captures `filter_set` via `nonlocal`. After registering a hook, `init_filter_set(register)` is called to prepare fresh state for the next registration — it creates a new `FilterSet` and attaches new include/exclude closures to `register.apply_to/register.skip_for`. However, the return value was discarded:

  hook.filter_set = filter_set          # assigns current (stale) outer filter_set
  init_filter_set(register)             # creates new FilterSet locally, return value ignored

Because `init_filter_set` creates the new `FilterSet` as a local variable (no nonlocal `filter_set`), the outer `filter_set` that register references is never updated. The next `apply_to` call correctly populates the new `FilterSet` (via the replaced include closure), but `register` then assigns the old one to the hook.

#### Fix

One-liner: capture the return value so the outer `filter_set` stays in sync.

### Checklist

- [x] Added changelog entry
- [x] Updated README/documentation, if necessary (not needed)

### DIFF
--- a/test/hooks/test_filters.py
+++ b/test/hooks/test_filters.py
@@ -1,5 +1,3 @@
-from types import SimpleNamespace
-
 import pytest
 from hypothesis import given, settings
 from hypothesis import strategies as st
@@ -186,45 +184,33 @@ def test_filter_body_works_in_negative_mode(ctx):
     inner()
 
 
-def _make_hook_context(operation_id):
-    """Build a minimal HookContext whose operation carries the given operationId."""
-    operation = SimpleNamespace(
-        definition=SimpleNamespace(raw={"operationId": operation_id}),
-        label=operation_id,
-        method="POST",
-        path=f"/fake/{operation_id}",
-        tags=[],
-    )
-    return HookContext(operation=operation)
+def _build_operation(ctx, operation_id, path):
+    op = {"operationId": operation_id, "responses": {"200": {"description": "OK"}}}
+    schema = schemathesis.openapi.from_dict(ctx.openapi.build_schema({path: {"post": op}}))
+    return schema[path]["POST"]
 
 
+# Each hook registered with a different apply_to filter must receive its own filter_set
 def test_multiple_apply_to_have_distinct_filters(ctx):
-    """Hooks registered with different apply_to filters must each receive their own filter_set.
-
-    Regression: a bug in to_filterable_hook caused the outer ``filter_set``
-    variable to go stale after the first registration, so every subsequent hook
-    was silently assigned the *first* hook's filter_set.
-    """
     with ctx.restore_hooks():
 
         @schemathesis.hook.apply_to(operation_id="operationA")
         def before_call(context, case, **kwargs):
-            """Hook for operation A."""
+            pass
 
         hook_a = before_call
 
         @schemathesis.hook.apply_to(operation_id="operationB")
         def before_call(context, case, **kwargs):  # noqa: F811
-            """Hook for operation B."""
+            pass
 
         hook_b = before_call
 
-        # Each hook must carry its own, distinct filter_set
         assert hook_a.filter_set is not hook_b.filter_set
 
-        ctx_a = _make_hook_context("operationA")
-        ctx_b = _make_hook_context("operationB")
-        ctx_other = _make_hook_context("operationC")
+        ctx_a = HookContext(operation=_build_operation(ctx, "operationA", "/a"))
+        ctx_b = HookContext(operation=_build_operation(ctx, "operationB", "/b"))
+        ctx_other = HookContext(operation=_build_operation(ctx, "operationC", "/c"))
 
         # hook_a should run for operationA only
         assert not _should_skip_hook(hook_a, ctx_a)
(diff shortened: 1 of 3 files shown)


test/hooks/test_filters.py
```diff
@@ -181,3 +184,85 @@ def inner(case):
...
+def _make_hook_context(operation_id):
+    """Build a minimal HookContext whose operation carries the given operationId."""
+    operation = SimpleNamespace(
+        definition=SimpleNamespace(raw={"operationId": operation_id}),
+        label=operation_id,
+        method="POST",
+        path=f"/fake/{operation_id}",
+        tags=[],
+    )
+    return HookContext(operation=operation)
```
Stranger6667: I'd prefer not using mock-like approach in tests at all where reasonably possible. There are already existing fixtures/helpers that can build `APIOperation` instance, i.e. there is a test "Schema" instance, so you can do `schema["/whatever"]["post"]`. then the whole `_make_hook_context` won't be needed

test/hooks/test_filters.py
```diff
@@ -181,3 +184,85 @@ def inner(case):
...
+
+
+def test_multiple_apply_to_have_distinct_filters(ctx):
+    """Hooks registered with different apply_to filters must each receive their own filter_set.
+
+    Regression: a bug in to_filterable_hook caused the outer ``filter_set``
+    variable to go stale after the first registration, so every subsequent hook
+    was silently assigned the *first* hook's filter_set.
+    """
```
Stranger6667: Please, follow existing test structure in this repo - no docstring, but rather a short and concise comment.

> Regression: a bug in to_filterable_hook caused the outer ``filter_set``

this kind of wording just pollutes the space with implementation details that could go stale very soon, i.e. function names, etc is not really important. User-facing behavior is much more important to note

test/hooks/test_filters.py
```diff
@@ -181,3 +184,85 @@ def inner(case):
...
+
+        @schemathesis.hook.apply_to(operation_id="operationA")
+        def before_call(context, case, **kwargs):
+            """Hook for operation A."""
+
+        hook_a = before_call
+
+        @schemathesis.hook.apply_to(operation_id="operationB")
+        def before_call(context, case, **kwargs):  # noqa: F811
+            """Hook for operation B."""
```
Stranger6667: this does not bring any value, just use `pass`

test/hooks/test_filters.py
```diff
@@ -181,3 +184,85 @@ def inner(case):
...
+
+    Regression: a bug in to_filterable_hook caused the outer ``filter_set``
+    variable to go stale after the first registration, so every subsequent hook
+    was silently assigned the *first* hook's filter_set.
+    """
+    with ctx.restore_hooks():
+
+        @schemathesis.hook.apply_to(operation_id="operationA")
+        def before_call(context, case, **kwargs):
+            """Hook for operation A."""
```
Stranger6667: this does not bring any value, just use `pass`

test/hooks/test_filters.py
```diff
@@ -181,3 +184,85 @@ def inner(case):
...
+        assert _should_skip_hook(hook_b, ctx_other)
+
+
+def test_multiple_apply_to_with_skip_for(ctx):
+    """Mixing apply_to and skip_for across multiple hooks must keep filters independent."""
+    with ctx.restore_hooks():
+
+        @schemathesis.hook.apply_to(operation_id="opInclude")
+        def before_call(context, case, **kwargs):
+            """Include-filtered hook."""
```
Stranger6667: this does not bring any value, just use `pass`

test/hooks/test_filters.py
```diff
@@ -181,3 +184,85 @@ def inner(case):
...
+
+        hook_include = before_call
+
+        @schemathesis.hook.skip_for(operation_id="opExclude")
+        def before_call(context, case, **kwargs):  # noqa: F811
+            """Exclude-filtered hook."""
```
Stranger6667: this does not bring any value, just use `pass`

test/hooks/test_filters.py
```diff
@@ -181,3 +184,85 @@ def inner(case):
...
+        assert _should_skip_hook(hook_a, ctx_other)
+
+        # hook_b should run for operationB only
+        assert _should_skip_hook(hook_b, ctx_a)
+        assert not _should_skip_hook(hook_b, ctx_b)
+        assert _should_skip_hook(hook_b, ctx_other)
+
+
+def test_multiple_apply_to_with_skip_for(ctx):
+    """Mixing apply_to and skip_for across multiple hooks must keep filters independent."""
```
Stranger6667: Use a comment for it

Stranger6667 (changes requested):
Great! Thanks - I've added a few notes on tests to make them in line with the rest of the codebase

Stranger6667 (approved):

PR #3631, opened 2026-03-20
fix(corpus): switch to safe_load in tools

tejasae-afk:
Was going through corpus/tools.py and this caught my eye.

**What's going on:** yaml.load() without a safe loader can instantiate arbitrary Python objects from the YAML content. Switched to safe_load() which only allows basic types.

**Why it matters:** This is a security concern that could be exploited in production.

Feel free to tweak the approach if needed.


Stranger6667:
Sure. though

> This is a security concern that could be exploited in production.

doesn't really hold for corpus tests. I.e. there is no "production" as it is not a part of the distribution

tejas-ae:
Makes sense. Thanks!

PR #3635, opened 2026-03-21
fix: Validate length constraints after regex-based string generation …

radionest:
Description

Update_quantifier cannot encode minLength/maxLength into regex quantifiers for patterns with bare character classes (IN opcode), e.g. ^[a-z][a-z0-9]*(-[a-z0-9]+)*$ with minLength=3. The pattern is returned unchanged, so the coverage phase generates strings like "a" that violate the length constraint.

This adds .filter() on the st.from_regex() strategy to enforce minLength/maxLength after generation — the same approach hypothesis-jsonschema uses internally. update_quantifier is kept as a best-effort optimization to reduce the filter rejection rate. 

Checklist

[x] Added failing tests for the change
[x] All new and existing tests pass
[x] Added changelog entry
[ ] Updated README/documentation, if necessary

### DIFF
--- a/src/schemathesis/generation/coverage.py
+++ b/src/schemathesis/generation/coverage.py
@@ -402,11 +402,16 @@ class CoverageContext:
                 re.compile(pattern)
             except re.error:
                 raise Unsatisfiable from None
-            if "minLength" in schema or "maxLength" in schema:
-                min_length = schema.get("minLength")
-                max_length = schema.get("maxLength")
+            min_length = schema.get("minLength")
+            max_length = schema.get("maxLength")
+            if min_length is not None or max_length is not None:
                 pattern = update_quantifier(pattern, min_length, max_length)
-            return cached_draw(st.from_regex(pattern))
+            strategy = st.from_regex(pattern)
+            if min_length is not None:
+                strategy = strategy.filter(lambda s: len(s) >= min_length)
+            if max_length is not None:
+                strategy = strategy.filter(lambda s: len(s) <= max_length)
+            return cached_draw(strategy)
         if (keys == ["items", "type"] or keys == ["items", "minItems", "type"]) and isinstance(schema["items"], dict):
             items = schema["items"]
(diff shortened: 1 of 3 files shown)


src/schemathesis/generation/coverage.py
```diff
@@ -402,11 +402,16 @@ def generate_from_schema(self, schema: JsonSchema) -> Any:
...
+            min_length = schema.get("minLength")
+            max_length = schema.get("maxLength")
+            if min_length is not None or max_length is not None:
                 pattern = update_quantifier(pattern, min_length, max_length)
-            return cached_draw(st.from_regex(pattern))
+            strategy = st.from_regex(pattern)
+            if min_length is not None:
+                strategy = strategy.filter(lambda s: len(s) >= min_length)
+            if max_length is not None:
+                strategy = strategy.filter(lambda s: len(s) <= max_length)
```
Stranger6667: I think we can encode 3 cases, so we have 1 lambda when min_length & max_length are both present

Stranger6667 (commented):
Good catch! I have a minor perf-related comment (which is surely negligible anyway)

radionest:
Sorry I didn't catch this earlier — the interaction between the filter, hypothesis generators and `_positive_string` turned out to be trickier than it looked at first glance.

The `.filter()` I added was doing two unexpected things: killing the entire `_positive_string` generator when an impossible boundary case raised `Unsatisfiable` (so valid boundaries like maxLength=60 were never reached), and forcing `st.from_regex` to produce garbage strings through partial matching when it couldn't find a value of the right length.

Working on a fix now — `fullmatch=True` to stop the garbage, and `try/except` around individual boundary cases so one failure doesn't take down the rest.

Stranger6667:
@radionest I've addressed these concerns in https://github.com/schemathesis/schemathesis/pull/3635/commits/c436308140971ec202d134a0fa793058b6b1610c

radionest:
Noticed it right when I was pushing my)

PR #3689, opened 2026-04-03
Ast rewrite clean

radionest:
Description

Rewrite update_quantifier from string surgery to the AST pipeline: parse → transform → serialize. This fixes edge cases with multi-char group inner lengths (e.g. (\d{3})+) and max_length=0 handling, while making the code easier to extend for future pattern support.

Checklist
 
+ Added failing tests for the change
+ All new and existing tests pass
- Added changelog entry
- Updated README/documentation, if necessary

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
       - id: yamllint
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.19.1
+    rev: v1.20.0
     hooks:
       - id: mypy
         exclude: ^(docs/|test/|test-corpus/|example/|benches/).*$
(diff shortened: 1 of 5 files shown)


Stranger6667 (approved):
Amazing work! thank you. I see there are a few things not covered by tests, but that's fine - I'll think about a few extra test cases. One thought I had about the branch you've shared earlier is that I'd expect some of those tests to run too long for a regular CI run - right now it does not feel super necessary given test cases you've already added here. But we can discuss it surely

radionest:
Thanks a lot, really glad to hear that! I agree that most of the tests from the first branch are not realy needed in CI -- pulling Rust just for testing is probably an overkill, and they would take too long anyway. I mostly needed them to make sure all edge cases are covered, and to make it easyer to verify that everything works correctly.

But the serializer doesn't cover the full regex syntax yet (e.g. lookahead/lookbehind), so if something needs to be changed there later, there should be a way to catch all edge cases. I'm not good enough with regex to reason about all of them in my head, so the most reliable approach I see is oracle-based testing. Maybe the corpus could be trimmed down to a smaller representative subset in the future.

PR #3721, opened 2026-04-16
docs: add type annotations to `guides/auth.md`

sobolevn:
It was really hard for me to figure out the right types for everything :)

### DIFF
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -137,14 +137,14 @@ import schemathesis
 
 @schemathesis.auth()
 class TokenAuth:
-    def get(self, case, ctx):
+    def get(self, case: schemathesis.Case, ctx: schemathesis.AuthContext) -> str:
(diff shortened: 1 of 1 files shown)


Stranger6667:
Thank you!

docs/guides/auth.md
```diff
@@ -156,16 +156,16 @@ Schemathesis caches tokens for 300 seconds by default.
 ```python
 @schemathesis.auth(refresh_interval=600)  # Refresh every 10 minutes
 class RefreshableAuth:
-    def __init__(self):
+    def __init__(self) -> None:
         self.refresh_token = None
 
-    def get(self, case, ctx):
+    def get(self, case: case.Case, ctx: schemathesis.AuthContext) -> str:
```
sobolevn, line 162: Oups, this is a wrong copy-paste :)

PR #3740, opened 2026-04-17
feat(junit): add group-by option for phase-based test suite separation

### LINKED ISSUE #3738: JUnit report marks all test cases as skipped when schema has no examples
## Description

When using JUnit reporting with an OpenAPI schema that has no examples, **all** test cases in the JUnit report are marked as skipped:

```xml
<skipped type="skipped">No examples in schema</skipped>
```

This occurs even though the coverage and fuzzing phases run successfully and produce results — the coverage report shows actual test execution, but the JUnit XML contradicts it.

## Root Cause

The `JunitXmlWriter` maintains a single `TestCase` per operation label across all phases (`_test_cases: dict[str, TestCase]`). When phases execute in order:

1. **Examples phase** — no examples in schema → `SkipTest` → `add_skipped_info("No examples in schema")` is called on the `TestCase`
2. **Coverage/Fuzzing phases** — tests run successfully → elapsed time is added to the **same** `TestCase`, but the `<skipped>` element is never cleared

Since `junit-xml`'s `is_skipped()` returns `True` whenever the `skipped` list is non-empty, CI tools treat every test case as skipped regardless of subsequent phase results.

### Relevant code paths

- `src/schemathesis/engine/run/unit/_executor.py:141-145` — translates `SkipTest` to `"No examples in schema"`
- `src/schemathesis/reporting/junitxml.py:29,40-56` — shared `_test_cases` dict and `record_scenario` logic
- `src/schemathesis/cli/commands/run/handlers/junitxml.py:27-48` — handler dispatches all phases to the same writer

## Reproduction

Configuration (no `[phases.examples] enabled = false`):

```toml
[reports.junit]
enabled = true

[generation]
max-examples = 50

[phases.stateful]
enabled = false

[phases.coverage]
unexpected-methods = []
```

Run against any OpenAPI schema **without** inline examples. Every entry in the resulting JUnit XML will contain `<skipped>`.

**Workaround**: disable the examples phase explicitly:

```toml
[phases.examples]
enabled = false
```

## Suggested Implementation

Add a `group-by` option to the JUnit report configuration that allows separating results by phase into distinct JUnit test suites, avoiding the state-mixing issue.

### Config surface

```toml
[reports.junit]
enabled = true
group-by = "phase"    # default: "operation" (current behavior, backward-compatible)
```

- `"operation"` (default): current behavior — one test suite, one test case per operation
- `"phase"`: one `<testsuite>` per phase (Examples, Coverage, Fuzzing), each with independent test cases

### Changes required

1. **`config/schema.json`**: add a `JunitReportConfig` definition extending `ReportConfig` with `"group-by": {"enum": ["operation", "phase"]}`
2. **`config/_report.py`**: add `JunitReportConfig` dataclass with a `group_by` field (default `"operation"`)
3. **`cli/executor.py`**: pass `JunitReportConfig` to the handler constructor
4. **`cli/commands/run/handlers/junitxml.py`**: forward the phase from events to the writer
5. **`reporting/junitxml.py`**: when `group_by="phase"`, key test cases by `(phase, label)` and emit one `TestSuite` per phase in `close()`

All the required phase information is already available on the engine events (`ScenarioFinished.phase`, `NonFatalError.phase`, `FuzzScenarioFinished` is implicitly Fuzzing).

baywet:
## Summary

Add a `group-by` configuration option to JUnit XML reports that controls how test cases are organized into test suites.

### Problem

When an OpenAPI schema has no inline examples, the examples phase marks all test cases as `<skipped>` in the JUnit report. Because test cases are shared across phases in a flat dictionary, this skip status persists even after the coverage and fuzzing phases run successfully, resulting in all entries showing as skipped.

### Solution

- **`group-by = "operation"`** (default): Test cases grouped by API operation with improved skip-clearing logic. When a non-skip result (success, failure, error) arrives from any phase, prior skip markers are cleared, fixing the bug while maintaining backward compatibility.

- **`group-by = "phase"`**: Each testing phase (Examples, Coverage, Fuzzing) gets its own `<testsuite>` element, providing clear per-phase visibility in CI/CD systems.

### Configuration

```toml
[reports.junit]
group-by = "phase"
```

### Changes

- Config layer: `JunitGroupBy` enum, `JunitReportConfig` dataclass, JSON schema
- Writer layer: Dual-mode `JunitXmlWriter` with skip-clearing logic and per-phase suites
- Handler layer: Phase info forwarding from engine events to writer
- 16 unit tests + 4 integration tests
- Documentation updates across 4 files

Fixes schemathesis/schemathesis#3738

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -243,6 +243,45 @@ parameters = { "path.user_id" = 42, "query.user_id" = 100 }
     path = "./test-reports/schemathesis-results.xml"
     ```
 
+#### `reports.junit.group-by`
+
+!!! note ""
+
+    **Type**: `String`  
+    **Default**: `"operation"`  
+    **Possible values**: `"operation"`, `"phase"`  
+
+    Controls how test cases are grouped in the JUnit XML report.
+
+    - **`operation`** (default): All phases contribute to a single test case per API operation. A single `<testsuite>` is generated.
+    - **`phase`**: Each execution phase (Examples, Coverage, Fuzzing) produces its own `<testsuite>`, with independent test cases per operation.
+
+    Phase grouping is useful when your schema has no inline examples — the Examples phase would mark test cases as skipped, while Coverage and Fuzzing phases run normally. With the default `operation` mode, skip results from one phase are automatically cleared when a later phase produces real results. With `phase` mode, each phase is fully isolated:
+
+    ```toml
+    [reports.junit]
+    enabled = true
+    group-by = "phase"
+    ```
+
+    **Example output with `group-by = "phase"`:**
+
+    ```xml
+    <testsuites>
+      <testsuite name="schemathesis - Examples">
+        <testcase name="GET /users">
+          <skipped type="skipped">No examples in schema</skipped>
+        </testcase>
+      </testsuite>
+      <testsuite name="schemathesis - Coverage">
+        <testcase name="GET /users" time="0.5"/>
+      </testsuite>
+      <testsuite name="schemathesis - Fuzzing">
+        <testcase name="GET /users" time="1.2"/>
+      </testsuite>
+    </testsuites>
+    ```
+
 ### Output
 
 #### `output.sanitization.enabled`
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -243,7 +243,7 @@ parameters = { "path.user_id" = 42, "query.user_id" = 100 }
     path = "./test-reports/schemathesis-results.xml"
     ```
 
-#### `reports.junit.group-by`
+#### `reports.group-by`
 
 !!! note ""
 
@@ -251,7 +251,7 @@ parameters = { "path.user_id" = 42, "query.user_id" = 100 }
     **Default**: `"operation"`  
     **Possible values**: `"operation"`, `"phase"`  
 
-    Controls how test cases are grouped in the JUnit XML report.
+    Controls how test cases are grouped in reports (currently applies to JUnit XML).
 
     - **`operation`** (default): All phases contribute to a single test case per API operation. A single `<testsuite>` is generated.
     - **`phase`**: Each execution phase (Examples, Coverage, Fuzzing) produces its own `<testsuite>`, with independent test cases per operation.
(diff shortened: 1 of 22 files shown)


docs/reference/configuration.md
```diff
@@ -243,6 +243,45 @@ parameters = { "path.user_id" = 42, "query.user_id" = 100 }
...
+    **Type**: `String`  
+    **Default**: `"operation"`  
+    **Possible values**: `"operation"`, `"phase"`  
+
+    Controls how test cases are grouped in the JUnit XML report.
+
+    - **`operation`** (default): All phases contribute to a single test case per API operation. A single `<testsuite>` is generated.
+    - **`phase`**: Each execution phase (Examples, Coverage, Fuzzing) produces its own `<testsuite>`, with independent test cases per operation.
+
+    Phase grouping is useful when your schema has no inline examples — the Examples phase would mark test cases as skipped, while Coverage and Fuzzing phases run normally. With the default `operation` mode, skip results from one phase are automatically cleared when a later phase produces real results. With `phase` mode, each phase is fully isolated:
```
Stranger6667, line 259: I am not sure if it is the right approach because there was already logic for "upgrading" per operation statuses, so changing grouping looks like hiding the incomplete/buggy status upgrades

Stranger6667: Ie it should happen on the engine/ctx level so it applies to all report types, not only to junit

baywet: I couldn't come up with alternative strategies beyond:
- manual script on my end to somehow remove the skipped inserted by the missing examples.
- grouping (this PR) by phase, so the ingestion process can look at only the test suites I care about (the execution)
- having the example phase stop emitting skipped entries (but that'd probably be breaking for a bunch of people)
- having some kind of "filtering" configuration for what gets exported to the junit report (which phase), but this implementation does not prevent that other aspect.

Thoughts? Better ideas?

baywet: just so your second message out of sequence. So you're effectively not pushing back against grouping, just the setting is at the wrong level. Got it. Let me push an update. Would you also expect all the reporters to implement that new setting at the same time? or just junit for now is fine?

Stranger6667:
Superseded by #3966

docs/reference/configuration.md
Stranger6667: Sorry for the delay! I've implemented it similarly to how the output handling does it - check via `ctx` about what actually has happened and then not reporting "skipped" in the case if wasn't truly skipped

PR #4162, opened 2026-05-18
perf: More caching

Stranger6667:


Stranger6667:

PR #4164, opened 2026-05-18
perf: Avoid hashing in some cases

Stranger6667:


Stranger6667:

PR #4178, opened 2026-05-22
perf: ~50% lower coverage-phase wall on definition-heavy schemas

Stranger6667:


Stranger6667:

PR #4215, opened 2026-05-30
perf: Lower memory usage in long stateful runs

Stranger6667:


Stranger6667:

PR #4234, opened 2026-06-08
Fix session cookie leakage between generated requests

HennieLP:
### Description

This change prevents cookies set by one response from being carried into later generated requests when Schemathesis reuses an engine-managed requests.Session. That stops auth cookies like session from accidentally authenticating follow-up cases and causing false positives such as “API accepts invalid authentication.”
### Checklist

- [x] Added changelog entry
- [x] Updated README/documentation, if necessary


Stranger6667:
Hi! Thanks, I'll push a few fixes & merge it

PR #4266, opened 2026-06-23
refactor: Drop the unnecessary `httpx` dependency

Stranger6667:


Stranger6667:
@codspeedbot fix this regression

PR #4417, opened 2026-08-04
fix: key parameter bundle cache on resolved reference

### LINKED ISSUE #4414: [BUG] Parameter bundle cache keyed on id() of freed dicts — intermittent "Path parameter is not defined" on multi-file specs
### Checklist

- [x] I checked the [[FAQ section](https://schemathesis.readthedocs.io/en/stable/faq/)](https://schemathesis.readthedocs.io/en/stable/faq/) of the documentation
- [x] I looked for similar issues in the [[issue tracker](https://github.com/schemathesis/schemathesis/issues)](https://github.com/schemathesis/schemathesis/issues)
- [x] I am using the latest version of Schemathesis

### Describe the bug
On a multi-file OpenAPI spec, `schemathesis run` intermittently fails with

```
Schema Error: Path parameter 'xxx' is not defined
```

The runs are against the same spec, with a pinned `--seed`, and against the same server. The parameter named in the error is defined in the spec. From one run to the next, schemathesis flags different failures.
### To Reproduce
It seems like the underlying issue is incorrect cache behavior. I'm attaching a script that produces a schema and then uses schemathesis to parse the schema and then validate the cache.

Run the attached repro.py using `uv run repro.py`. This script creates the schema in a directory created in the current location. The schema creates a large number of trivial GET operations that all return 200 and then calls calls `schema.get_all_operations()` to exercise schema parsing for multiple rounds.

The output (below) identifies path parameters injected by schemathesis even though the parameter is defined in the schema.

```
% uv run repro.py
python 3.14.0  schemathesis 4.24.3
external path-item files: 200   KEEPALIVE=False

  round 1: bundle_cache=  200  operations_with_injected_placeholder=0
  round 2: bundle_cache=  400  operations_with_injected_placeholder=0
  round 3: bundle_cache=  591  operations_with_injected_placeholder=9
            GET /items/145/{op145_id}
              lost 'op145_id' -> placeholder; gained 'op1_id' from another operation
            GET /items/147/{op147_id}
              lost 'op147_id' -> placeholder; gained 'op87_id' from another operation
            GET /items/149/{op149_id}
              lost 'op149_id' -> placeholder; gained 'op89_id' from another operation
  round 4: bundle_cache=  741  operations_with_injected_placeholder=50
            GET /items/80/{op80_id}
              lost 'op80_id' -> placeholder; gained 'op140_id' from another operation
            GET /items/81/{op81_id}
              lost 'op81_id' -> placeholder; gained 'op132_id' from another operation
            GET /items/82/{op82_id}
              lost 'op82_id' -> placeholder; gained 'op22_id' from another operation
  round 5: bundle_cache=  878  operations_with_injected_placeholder=63
            GET /items/26/{op26_id}
              lost 'op26_id' -> placeholder; gained 'op36_id' from another operation
            GET /items/28/{op28_id}
              lost 'op28_id' -> placeholder; gained 'op168_id' from another operation
            GET /items/30/{op30_id}
              lost 'op30_id' -> placeholder; gained 'op170_id' from another operation
  round 6: bundle_cache= 1012  operations_with_injected_placeholder=66
            GET /items/15/{op15_id}
              lost 'op15_id' -> placeholder; gained 'op118_id' from another operation
            GET /items/17/{op17_id}
              lost 'op17_id' -> placeholder; gained 'op157_id' from another operation
            GET /items/19/{op19_id}
              lost 'op19_id' -> placeholder; gained 'op159_id' from another operation

bundle cache entries after 6 parses: 1012  (spec only has 200 parameters)
entries missing vs 6 x 200 = 1200: 188
cache hits that returned a DIFFERENT parameter's definition: 188
  round 3: id=0x10b507c80 was cached in round 1 for ('op1_id', 'path') but was hit by ('op145_id', 'path')
  round 3: id=0x10b5bfc00 was cached in round 2 for ('op87_id', 'path') but was hit by ('op147_id', 'path')
  round 3: id=0x10b5c4380 was cached in round 2 for ('op89_id', 'path') but was hit by ('op149_id', 'path')
  round 3: id=0x10b529300 was cached in round 1 for ('op147_id', 'path') but was hit by ('op154_id', 'path')
  round 3: id=0x10b593b00 was cached in round 1 for ('op95_id', 'path') but was hit by ('op155_id', 'path')

total spurious injected path parameters: 188
```

**The script generates the schema**, but I am also attaching it for reference. Note: the minimal schema is fairly large. The number of files are due to needing to exceed the cache capacity that triggers the issue.
### Expected behavior
The expected behavior is that the parameters defined on endpoints are consistent. More specifically, the cache should provide a valid parameter definition on a hit, or a miss if the value is not currently stored in the cache.
### Environment

```
- OS: MacOS 26.6
- Python version: Python 3.14.0
- Schemathesis version: 4.24.3
- Spec version: `openapi: 3.0.3`
```

### Additional context
The script includes ENV flags that drive behavior. KEEPALIVE ensures that dicts are never recycled and the issue gets resolved -- to be clear, this is not a solution I am proposing. The FILES count must be > 128 for the issue to appear - this is the lru_cache_limit. The call to `churn` exacerbates the problem, but can be disabled using NOCHURN. Finally, we use `st_parameters._bundle_parameter()` to wrap the original call for instrumentation. This can be disabled using NOINSTRUMENT to confirm the issue is not introduced by the instrumentation.

I do plan to include a PR and will link to this issue for context.

[repro.py](https://github.com/user-attachments/files/30709420/repro.py)
[spec.zip](https://github.com/user-attachments/files/30709431/spec.zip)

davidmarkcarlson:
Fixes [#4414](https://github.com/schemathesis/schemathesis/issues/4414)

`_bundle_parameter` cached bundled parameter definitions in a dict keyed on
`id(parameter)` while keeping no reference to the dicts it keyed on. The cache
lives on the schema and outlives the parse that filled it, but parameter dicts
reached through a `$ref` are rebuilt on every parse and freed straight after.
CPython then reuses those addresses, a later parameter lands on a stale entry,
and the cache returns a different parameter's definition.

The affected operation loses its own path parameter. schemathesis creates
a placeholder injected in its place and adds this parameter to an unrelated
route. As a result, `create_test` reports `Path parameter '<name>' is not defined`
against a schema that declares it correctly. Because it is driven by allocator
behaviour, it varies run to run on a byte-identical schema with a pinned seed.

This keys on the parameter's resolved `$ref` URI instead — the definition itself 
rather than the object citing it. `resolve_reference_uri` already computes and 
memoises it, so the key is a cheap string. `maybe_resolve_with_resolver` follows 
the `$ref` chain and discards siblings, so the bundled result depends only on that 
URI. Inline parameters are not cached: they are unique to their operation, so there 
is nothing to share, and hashing their contents would cost more than the bundling 
it saves.

**It also makes the cache work.** `_bundle_parameter` is called with the citation.
The `{"$ref": ...}` dict inside each path item — not the definition it resolves to. 
Every citation is a distinct object, so twenty operations referencing one parameter 
produced twenty entries and no hits. Measured over 6 parses of a 68-operation 
schema (408 calls)

| key | hits |
|---|---|
| `id()` | 22 (5.4%), and most of those were the bug |
| resolved `$ref` URI | 300 (73.5%), all correct |

Parsing that schema goes from 0.122s to 0.056s over 6 parses. Also, the cache 
stops growing. For this test case, we had produced 379 entries after 6 parses. 
Now we maintain a stable 14 entries.

### Verification

- The reproduction in #4414 goes from 149–316 spurious injected path parameters to **0**.
- Parse output is byte-identical to running with the cache disabled entirely, across 3 parses of the schema that surfaced this.
- Full suite test results): **9740 passed 5 skipped, 3 xfailed, 378 warnings**. These are the same skipped and expected failures as HEAD of master. 

### Checklist

- [x] Added changelog entry
- [x] ~~Updated README/documentation, if necessary~~ no user-facing API change

### Note on the test

`test_referenced_parameters_share_one_bundle_cache_entry` asserts the invariant directly rather than trying to provoke address reuse, so it is deterministic and will not flake in CI. On `master` it fails with output that states the problem by itself — two integer keys holding identical values:

```
{4503752704: ({'name': 'item_id', 'in': 'path', 'required': True, 'schema': {'type': 'string'}}, {}),
 4503656832: ({'name': 'item_id', 'in': 'path', 'required': True, 'schema': {'type': 'string'}}, {})}
```

It sits directly below `test_non_string_parameter_location`, which already covers the `isinstance(location, str)` branch this bug exploited to make parameters disappear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,12 @@
 - Runtime error in `negative_data_rejection` for empty array path parameters.
 - Coverage phase generating Unicode characters where the validator reads `\d` and `\w` as ASCII.
 - Internal error on truncated runtime expressions like `$response.` in Open API links.
+- Intermittent `Path parameter is not defined` errors on multi-file schemas, caused by the parameter bundle cache being keyed on object identity. [#4414](https://github.com/schemathesis/schemathesis/issues/4414)
 
 ### :racing_car: Performance
 
 - Cache canonical-form strategies across operations.
+- Share bundled parameters between operations that reference the same definition.
 
 ### :wrench: Changed
 
--- a/src/schemathesis/core/jsonschema/bundler.py
+++ b/src/schemathesis/core/jsonschema/bundler.py
@@ -13,8 +13,8 @@ from schemathesis.core.transforms import decode_pointer
 
 BUNDLE_STORAGE_KEY = "x-bundled"
 REFERENCE_TO_BUNDLE_PREFIX = f"#/{BUNDLE_STORAGE_KEY}"
-# Cache for bundled parameters: parameter object id -> (bundled definition, name_to_uri mapping)
-BundleCache = dict[int, tuple[dict[str, Any], dict[str, str]]]
+# Cache for bundled parameters: resolved `$ref` URI -> (bundled definition, name_to_uri mapping)
+BundleCache = dict[str, tuple[dict[str, Any], dict[str, str]]]
 
 
 class BundleError(Exception):
--- a/src/schemathesis/specs/openapi/adapter/parameters.py
+++ b/src/schemathesis/specs/openapi/adapter/parameters.py
@@ -25,7 +25,7 @@ from schemathesis.core.jsonschema import (
     schema_with_bundle,
 )
 from schemathesis.core.jsonschema.bundler import BUNDLE_STORAGE_KEY, BundleCache
-from schemathesis.core.jsonschema.resolver import Resolver
+from schemathesis.core.jsonschema.resolver import Resolver, resolve_reference_uri
 from schemathesis.core.jsonschema.types import JsonSchema, JsonSchemaObject, JsonValue, get_type
 from schemathesis.core.media_types import FORM_MEDIA_TYPES
 from schemathesis.core.parameters import HEADER_LOCATIONS, ParameterLocation
@@ -1556,13 +1556,17 @@ def _bundle_parameter(
     parameter: Mapping,
     resolver: Resolver,
     bundler: Bundler,
-    bundle_cache: dict[int, tuple[dict[str, Any], dict[str, str]]],
+    bundle_cache: BundleCache,
 ) -> tuple[dict[str, Any], dict[str, str]]:
     """Bundle a parameter definition to make it self-contained."""
-    param_id = id(parameter)
-    if param_id in bundle_cache:
-        cached_definition, cached_name_to_uri = bundle_cache[param_id]
-        return deepclone(cached_definition), dict(cached_name_to_uri)
+    # Key on the parameter's resolved `$ref` URI
+    reference = parameter.get("$ref")
+    cache_key = resolve_reference_uri(resolver.base_uri, reference) if isinstance(reference, str) else None
+    if cache_key is not None:
+        cached = bundle_cache.get(cache_key)
+        if cached is not None:
+            cached_definition, cached_name_to_uri = cached
+            return deepclone(cached_definition), dict(cached_name_to_uri)
 
     parameter_resolver, definition = maybe_resolve_with_resolver(parameter, resolver)
     schema = definition.get("schema")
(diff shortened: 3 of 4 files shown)


Stranger6667:
Sorry, the issue has been fixed already. See https://github.com/schemathesis/schemathesis/issues/4414#issuecomment-5181473801

davidmarkcarlson:
Thanks! I see it landed.

PR #4511, opened 2026-08-19
perf: Coverage phase up to 40x lower peak memory on operations with large deeply nested request bodies

Stranger6667:


Stranger6667:
@codspeedbot investigate this regression

PR #4526, opened 2026-08-21
chore: add httpx2 support

even-even:
Hello
In PR:
add https://github.com/pydantic/httpx2 (fork httpx by pydantic) support


src/schemathesis/core/transport.py
```diff
@@ -165,14 +165,14 @@ def from_any(cls, response: Response | httpx.Response | requests.Response | Test
...
         try:
-            import httpx
+            import httpx2
         except ImportError:
             pass
         else:
-            if isinstance(response, httpx.Response):
-                return Response.from_httpx(response, verify=True)
+            if isinstance(response, httpx2.Response):
+                return Response.from_httpx2(response, verify=True)
```
Stranger6667, line 185: This change implies a breakage for existing `httpx` users - we should support both, `httpx` and `httpx2` + same for typing. At least for some time

Stranger6667 (changes requested):

Stranger6667 (approved):

src/schemathesis/core/transport.py
even-even: Thanks for review.
1) return httpx
2) change title and description in PR
3) add changelog

PR #4576, opened 2026-08-27
chore: add SIM ruff rule

even-even:
### Description

1) add https://docs.astral.sh/ruff/rules/#flake8-simplify-sim ruff rule
2) fixes warns
3) add https://docs.astral.sh/ruff/settings/#show-fixes param in pre-commit

### Checklist

- [] Added failing tests for the change
- [ x] All new and existing tests pass
- [ ] Added changelog entry

### DIFF
--- a/src/schemathesis/transport/wsgi.py
+++ b/src/schemathesis/transport/wsgi.py
@@ -131,7 +131,9 @@ class WSGITransport(BaseTransport["werkzeug.Client"]):
             cookies=cookies,
         )
 
-        headers = {key: response.headers.getlist(key) for key in response.headers.keys()}
+        headers = {}
+        for name, value in response.headers:
+            headers.setdefault(name.lower(), []).append(value)
 
         return Response(
             status_code=response.status_code,
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -29,10 +29,10 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@ff2f1c621b7f889edc0d3c761ac2e6a3f8cdb0dd # v4.37.7
+        uses: github/codeql-action/init@db488ddef3bf6cb639b32c2e9a7c0a7ea8271d28 # v4.37.8
         with:
           languages: ${{ matrix.language }}
           build-mode: none
 
       - name: Perform CodeQL Analysis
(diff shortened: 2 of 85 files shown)


Stranger6667:
I don't think that opening PRs for refactorings like that without prior discussion is the best way to contribute. E.g. `suppress` is skipped deliberately in many places for performance reasons (yes, I've measured it - it matters in some places. `any` has a similar story) or I don't find ternary expressions particularly readable in all cases (and it collapses coverage reported for those lines). However, I am fine with similifications like for `.keys()` or single return instead of if + return True / return False, though I don't see them as particularly high value ones

even-even:
> I don't think that opening PRs for refactorings like that without prior discussion is the best way to contribute. E.g. `suppress` is skipped deliberately in many places for performance reasons (yes, I've measured it - it matters in some places. `any` has a similar story) or I don't find ternary expressions particularly readable in all cases (and it collapses coverage reported for those lines). However, I am fine with similifications like for `.keys()` or single return instead of if + return True / return False, though I don't see them as particularly high value ones

Hello
What do you think about in PR:
rollback
1) supress (and ignore in ruff config)
2) ternary (and ignore in ruff config)

leave it as is
1) with operator in open/read file
2) merge double if (if..if -> if .. and)
3) delete keys()

Stranger6667:
yep, it would work

even-even:
> > I don't think that opening PRs for refactorings like that without prior discussion is the best way to contribute. E.g. `suppress` is skipped deliberately in many places for performance reasons (yes, I've measured it - it matters in some places. `any` has a similar story) or I don't find ternary expressions particularly readable in all cases (and it collapses coverage reported for those lines). However, I am fine with similifications like for `.keys()` or single return instead of if + return True / return False, though I don't see them as particularly high value ones
> 
> Hello What do you think about in PR: rollback
> 
> 1. supress (and ignore in ruff config)
> 2. ternary (and ignore in ruff config)
> 
> leave it as is
> 
> 1. with operator in open/read file
> 2. merge double if (if..if -> if .. and)
> 3. delete keys()

done

src/schemathesis/transport/wsgi.py
```diff
@@ -131,7 +131,9 @@ def send(
             cookies=cookies,
         )
 
-        headers = {key: response.headers.getlist(key) for key in response.headers.keys()}
+        headers = {}
+        for name, value in response.headers:
+            headers.setdefault(name.lower(), []).append(value)
```
Stranger6667, line 136: why?

even-even: 1) https://docs.astral.sh/ruff/rules/in-dict-keys/
2) in my bench my solution work faster that old

Image

PR #4742, opened 2026-09-10
feat: Known-findings baseline so CI fails only on findings that are new

Stranger6667:


docs/guides/baseline.md
```diff
@@ -0,0 +1,141 @@
...
+schemathesis run http://localhost:8080/openapi.json --baseline-update
+```
+
+Add the file to `schemathesis.toml`:
+
+```toml
+baseline = "schemathesis-baseline.json"
+```
+
+Any filename works, and the path resolves from the directory you run in — not from wherever the config file was found. The run's header block echoes what it loaded, so a wrong path shows up before any request goes out:
```
Stranger6667: should be simplified. not really useful

docs/guides/baseline.md
```diff
@@ -0,0 +1,141 @@
...
+
+```
+Baseline:    schemathesis-baseline.json / 47 entries
+```
+
+Commit it. Every later run reports the recorded failures but exits `0`; anything else exits `1`.
```
Stranger6667: Is it a good practice? it has expiration dates

docs/guides/baseline.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+```
+Baseline:
+  Known failures: 47
+
+=========================== No issues found in 12.03s ==========================
+```
+
+The file changes only when you ask — a run without `--baseline-update` reads it and nothing more. The updating run itself still reports what it found, so leaving the flag in CI keeps the build red rather than quietly absorbing every regression.
```
Stranger6667: use simple "-"

docs/guides/baseline.md
```diff
@@ -0,0 +1,141 @@
...
+
+`Recorded` tells you when to stop — repeat until a run adds nothing:
+
+```
+Baseline:
+  Known failures: 37
+  Recorded: 0
+```
+
+Then commit the file. A run that still records on the tenth attempt means the identity is not settling — check whether those entries differ only in a status code that varies per request.
```
Stranger6667: Doesn't sound helpful - why don't we handle it automatically?

docs/guides/baseline.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Annotate entries
+
+Entries are yours to edit. `reason`, `ticket` and `expires` are preserved by every later `--baseline-update`, which stamps `first_seen` once and refreshes `last_seen` each time it sees the failure again:
```
Stranger6667: Why do we even hardcode fields that are not dropped? what if the user wants something else? why this arbitrary list of fields?

docs/guides/baseline.md
```diff
@@ -0,0 +1,141 @@
...
+  "id": "f3a91c",
+  "operation": "GET /users/{userId}",
+  "check": "positive_data_acceptance",
+  "failure": "RejectedPositiveData",
+  "signature": "409",
+  "first_seen": "2026-09-10",
+  "last_seen": "2026-09-14",
+  "reason": "Legacy conflict semantics on the v1 read path",
+  "ticket": "API-4412",
+  "expires": "2026-12-01"
```
Stranger6667: How is it decided when the issue expires?

docs/guides/baseline.md
```diff
@@ -0,0 +1,141 @@
...
+See the [CI/CD guide](cicd.md) for the surrounding setup.
+
+## What an entry matches
+
+A failure is identified by its operation, the check that raised it, the failure class, and one discriminating value — a status code, a JSON pointer into the schema, a header name. Generated payloads and case ids are deliberately excluded, so entries survive a change of seed.
+
+Two consequences worth knowing:
+
+- Some failure classes key on the status code alone, so one entry covers every failure of that class on that operation at that status. A second, genuinely different bug of the same shape there would be absorbed.
+- Custom checks are keyed partly by where in your code the assertion was raised, so their entries do not travel between machines with different checkout paths.
```
Stranger6667: Can't we unify it? sounds arbitrary - we can normalize paths against the project relative path + OS-wise

